PR #1563, opened 2024-07-26
CDD-2095: integrate headline manager to charts endpoint

phill-stanley:
# Description

This PR includes: Updates to the charts and plots interface to integrate support for `CoreHeadline` manager to enable charts to be generated from headline data.

Fixes #CDD-2095

---

## Type of change

---

# Checklist:


phill-stanley:
Screenshot 2024-07-26 at 11 41 26
Screenshot 2024-07-26 at 11 41 33

metrics/data/managers/core_models/headline.py
```diff
@@ -176,6 +176,74 @@ class CoreHeadlineManager(models.Manager):
     def get_queryset(self) -> CoreHeadlineQuerySet:
         return CoreHeadlineQuerySet(model=self.model, using=self.db)
 
+    def filter_for_x_and_y_values(
```
A-Ashiq, line 179: This same API that you were trying to mirror has since changed in the timeseries manager. Is this supposed to be `query_for_data()`?

metrics/domain/common/utils.py
```diff
@@ -91,3 +94,8 @@ def get_y_axis_value(cls, *, name: str) -> str:
 
 DEFAULT_X_AXIS = ChartAxisFields.get_default_x_axis().name
 DEFAULT_Y_AXIS = ChartAxisFields.get_default_y_axis().name
+
+
+class MetricTypes(Enum):
```
A-Ashiq, line 99: There is an enum called `DataSourceEnum` in the ingestion module.
Perhaps we could unify these? Maybe set a couple of class methods on there like so:

```
class MetricTypes(Enum):
    headline = "headline"
    cases = "cases"
    deaths = "deaths"
    healthcare = "healthcare"
    testing = "testing"
    vaccinations = "vaccinations"

     @property
     def is_headline(self):
            return self.value == "headline"
     
     @property
     def is_timeseries(self):
              return self.value != "headline"
``` 
Then we can just let the metric_group from the source file dictate that

metrics/domain/models/plots.py
```diff
@@ -84,7 +84,7 @@ def y_axis_value(self) -> str | None:
             return None
         return ChartAxisFields.get_y_axis_value(name=self.y_axis)
 
-    def to_dict_for_query(self) -> dict[str, str]:
+    def to_dict_for_query(self, metric_type: str) -> dict[str, str]:
```
A-Ashiq, line 87: If we follow my other suggestion, we can just infer this information from the `metric_group` which if we don't immediately have to hand, we can extract easily from the `metric`

metrics/interfaces/charts/access.py
```diff
@@ -38,18 +39,57 @@ def __init__(
...
+        Notes:
+            The charts interface can be used to generate charts
+            of either `Core timeseries` or `Core headline` data.
+            this function returns the type based on the provided
+            metric.
+
+        Returns:
+            a string of the metric type either `headline` or `timeseries`.
+        """
+        if self.chart_plots.plots[0].metric.split("_")[1] == MetricTypes.HEADLINE.value:
```
A-Ashiq, line 88: Yup so you've already kinda done it here. Let's move this further up the stack call so this contextual info is passed right the way down

metrics/domain/models/plots.py
phill-stanley: Updated this to with the `DataSourceEnum` change.

metrics/data/managers/core_models/headline.py
phill-stanley: I went back and forth a bit on this when I saw the update to timeseries but you're right. It should be aligned. I've made a few more changes to the interface as well to combine `get_headline_data` and `get_timeseries` now they're both using `query_for_data()`

ingestion/data_transfer_models/base.py
```diff
@@ -26,7 +27,7 @@ class IncomingBaseDataModel(BaseModel):
     parent_theme: enums.ParentTheme
     child_theme: str
     topic: str
-    metric_group: enums.DataSourceFileType
+    metric_group: DataSourceFileType
```
A-Ashiq, line 30: I think you should reference this enum via the interface class here as opposed to directly importing it from the file where the interface object lives. It will be convoluted but it makes sure that we funnel cross-boundary dependencies through the same points

metrics/domain/models/plots.py
```diff
@@ -35,6 +35,10 @@ def topic_name(self) -> str:
     def metric_name(self) -> str:
         return self.metric
 
+    @property
+    def metric_group(self) -> str:
+        return self.metric.split("_")[1]
```
A-Ashiq, line 40: This is worrying me now that we have more metrics coming into the system. We could very much have a topic name that has an `_` character in the name, which would throw this off.
I wonder if actually we should fetch the enum and loop through it, returning the value of the enum when it is found in the metric?

metrics/interfaces/charts/access.py
```diff
@@ -38,18 +39,39 @@ def __init__(
         self,
         *,
         chart_plots: PlotsCollection,
-        core_time_series_manager: Manager = DEFAULT_CORE_TIME_SERIES_MANAGER,
+        core_model_manager: Manager | None = None,
```
A-Ashiq, line 42: Maybe worth making a new type and that here:
```
CORE_MODEL_MANAGER_TYPE = CoreTimeSeries | CoreHeadline
```
So this would become:
```
core_model_manager: CORE_MODEL_MANAGER_TYPE | None = None,
```

metrics/interfaces/plots/access.py
```diff
@@ -70,16 +70,20 @@ def validate_plot_parameters(self) -> None:
                 raise InvalidPlotParametersError from error
 
     def get_queryset_result_for_plot_parameters(
-        self, *, plot_parameters: PlotParameters
-    ) -> QuerySetResult:
-        """Returns the timeseries records for the requested plot as an enriched `QuerySetResult` model.
+        self,
+        *,
+        plot_parameters: PlotParameters,
+    ):
```
A-Ashiq, line 76: Did you mean to remove the return type hint here?

tests/integration/metrics/data/managers/core_models/test_headline.py
```diff
@@ -9,6 +9,51 @@
...
 
 class TestCoreHeadlineManager:
+
+    @pytest.mark.django_db
+    def test_query_for_data_returns_expected_results(
+        self,
+    ):
+        """
+        Given a `CoreHeadline` record is live
+        When `filter_for_x_and_y_values()` is called
```
A-Ashiq, line 19: Should this be `query_for_data()`?

metrics/interfaces/plots/access.py
phill-stanley: No, that should still be there. I've added that back.

metrics/interfaces/plots/access.py
```diff
@@ -91,89 +95,30 @@ def get_queryset_result_for_plot_parameters(
...
+        queryset = self.get_core_api_data(plot_params=plot_params)
 
-        return QuerySetResult(
-            queryset=queryset,
-            latest_date=queryset.latest_date,
-        )
+        return QuerySetResult(queryset=queryset, latest_date=queryset.latest_date)
 
-    def get_timeseries(
+    def get_core_api_data(
```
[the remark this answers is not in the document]
phill-stanley: Yeah, good point. I'll rename this.

tests/integration/metrics/data/managers/core_models/test_headline.py
phill-stanley: Good catch, I've updated this.

metrics/domain/models/plots.py
phill-stanley: i'll take a look at this and make an update. If we think this could be an issue it will have an impact with the validation for the ingestion pipeline, which has some assumptions around the structure of metrics.

I'll add something to the back log to look at that.

metrics/domain/models/plots.py
```diff
@@ -99,20 +104,28 @@ def to_dict_for_query(self) -> dict[str, str]:
...
             "sex": self.sex or "",
             "age": self.age or "",
-            "date_from": self.date_from_value,
-            "date_to": self.date_to_value,
-            "x_axis": self.x_axis_value,
-            "y_axis": self.y_axis_value,
         }
 
+        if DataSourceFileType[self.metric_group].is_headline:
+            params["geography_code"] = ""
```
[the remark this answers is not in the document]
phill-stanley: Ah, this is a mistake to have this here. The `geography_code` isn't used by the endpoint but the manager supports it. It just defaults to and empty string but I'd put that here as well. Its not required.

metrics/data/managers/core_models/headline.py
```diff
@@ -176,6 +176,77 @@ class CoreHeadlineManager(models.Manager):
...
+        queryset = self.get_queryset().get_headlines_released_from_embargo(
+            topic_name=topic_name,
+            metric_name=metric_name,
+            geography_name=geography_name,
+            geography_type_name=geography_type_name,
+            geography_code=geography_code,
+            stratum_name=stratum_name,
+            age=age,
+            sex=sex,
+        )[:1]
```
A-Ashiq, line 238: What's your reasoning behind this? This is going to incur a database trip here so there's a penalty to be paid

metrics/domain/common/utils.py
```diff
@@ -91,3 +94,43 @@ def get_y_axis_value(cls, *, name: str) -> str:
...
+    """Returns the metric group based on the provided metric
+
+    Args:
+        metric: string representation a metric
+
+    Returns:
+        string of the metric group the provided metric belongs to
+        if no matching metric group is found then `None` is returned.
+    """
+    if DataSourceFileType.headline.value in metric:
```
A-Ashiq, line 129: I don't see why you need this? The loop below will catch `headline` as well?

metrics/domain/common/utils.py
```diff
@@ -91,3 +94,43 @@ def get_y_axis_value(cls, *, name: str) -> str:
...
+        return DataSourceFileType.headline.value
+
+    for metric_group in DataSourceFileType:
+        if metric_group.value in metric:
+            return metric_group.value
+
+    return None
```
A-Ashiq, line 136: I feel like raising some kind of error would be clearer and more explicit to the caller of this function

tests/unit/metrics/domain/common/test_utils.py
```diff
@@ -344,3 +354,106 @@ def test_returns_true_for_vaccination_testing_substrings(self, metric_name: str)
...
+        Given a valid metric with either a `Headline` or `Timeseries`
+            metric group
+        When the `is_timeseries` property is called
+        Then the expected response is returned
+        """
+        # Given
+        expected_metric_group = expected_return_value
+
+        # When
+        returned_metric_group = DataSourceFileType[metric_group].is_timeseries
```
A-Ashiq, line 421: Same comment as above

tests/unit/metrics/domain/common/test_utils.py
```diff
@@ -344,3 +354,106 @@ def test_returns_true_for_vaccination_testing_substrings(self, metric_name: str)
...
+        (
+            [
+                ("COVID-19_headline_vaccines_spring24Uptake", "headline"),
+                ("COVID-19_headline_cases_7DayPercentChange", "headline"),
+                ("COVID-19_cases_casesByDay", "cases"),
+                ("RSV_testing_positivityByWeek", "testing"),
+                ("COVID-19_deaths_ONSRegByWeek", "deaths"),
+                ("COVID-19_healthcare_occupiedBedsRollingMean", "healthcare"),
+                ("COVID-19_vaccinations_autumn23_uptakeByDay", "vaccinations"),
+                ("non-existent-metric", None),
```
A-Ashiq, line 439: Same comment as before, returning None feels a little opaque. We should raise an error instead so its clear to the caller of this function what has happened

metrics/domain/common/utils.py
phill-stanley: Good point, I'll update this and raise a `ValueError` for an invalid metric.

metrics/data/managers/core_models/headline.py
phill-stanley: I used a slice here because I wanted to return the queryset with the first item in it and using the .first() queryset method returned the headline item from within the queryset.

I've checked the docs and as long as we don't use the `step` parameter when slicing we won't incur an extra database call.

https://docs.djangoproject.com/en/5.0/ref/models/querysets/

tests/unit/metrics/domain/common/test_utils.py
```diff
@@ -344,3 +354,128 @@ def test_returns_true_for_vaccination_testing_substrings(self, metric_name: str)
...
+                "COVID-19_invalid_metric",
+            ]
+        ),
+    )
+    def test_extract_metric_raises_error(self, metric: str):
+        """
+        Given an invalid metric
+        When the `extract_metric_group_from_metric()` is called
+            with the metric value as its parameter
+        Then the expected `metric_group` is returned
```
A-Ashiq, line 474: Is it? Should this say the error was raised?

phill-stanley: Good catch, I've updated that to `a ValueError is raised.`

A-Ashiq (approved):

PR #1591, opened 2024-08-06
Bug/deprecated geographies migration

A-Ashiq:
# Description

- Adds a data migration to cover the case when the Mersey geography already exists in the db

Fixes #CDD-2134

---

## Type of change

---

# Checklist:

### DIFF
--- a/metrics/data/migrations/0028_data_migration_for_updated_nhs_geographies.py
+++ b/metrics/data/migrations/0028_data_migration_for_updated_nhs_geographies.py
@@ -96,6 +96,13 @@ def _update_nhs_trust(*, apps: StateApps) -> None:
         )
         return
 
+    if Geography.objects.filter(name=NHS_TRUST_UPDATE_LOOKUP["new_name"]).exists():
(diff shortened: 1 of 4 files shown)


metrics/data/migrations/0028_data_migration_for_updated_nhs_geographies.py
```diff
@@ -96,6 +96,13 @@ def _update_nhs_trust(*, apps: StateApps) -> None:
         )
         return
 
+    if Geography.objects.filter(name=NHS_TRUST_UPDATE_LOOKUP["new_name"]).exists():
```
A-Ashiq, line 99: We have to check for existence as opposed to trying to catch the error because the migration test helper will see the error within the call stakc and always try to rollback the transaction

A-Ashiq: This will mean the migration that is currently falling over can just pass through. And we'll rely on `0029` to pick up the baton from here

PR #1753, opened 2024-09-17
CDD-2195: Use `plot_label` when provided for headline reference field

phill-stanley:
# Description

This PR includes an update to the tables endpoint for `Headline` data, when a label is provided for a plot this label is used to the populate the `reference` field in place of the `x_axis` value. This allows for more control over the presentation of tabular data for charts that use none date values.

Fixes #CDD-2195

---

## Type of change

---

# Checklist:

### DIFF
--- a/metrics/domain/tables/generation.py
+++ b/metrics/domain/tables/generation.py
@@ -14,6 +14,8 @@ IN_REPORTING_DELAY_PERIOD = "in_reporting_delay_period"
 PLOT_DATA_LOOKUP_TYPE = dict[datetime.date, Decimal]
 IN_REPORTING_DELAY_PERIOD_LOOKUP_TYPE = dict[datetime.date, bool]
 
+DEFAULT_PLOT_LABEL = "Plot"
+
 
 class TabularData:
     def __init__(self, *, plots: list[PlotData]):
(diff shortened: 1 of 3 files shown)


phill-stanley:
Screenshot 2024-09-17 at 08 56 12
Screenshot 2024-09-17 at 08 56 21
Screenshot 2024-09-17 at 08 56 42
Screenshot 2024-09-17 at 08 57 07

metrics/domain/tables/generation.py
```diff
@@ -190,6 +197,11 @@ def create_multi_plot_row(
             )
 
         if DataSourceFileType[self.metric_group].is_headline:
+            current_plot_label = list(plot_values.keys())[0]
```
A-Ashiq, line 200: Nitpick: I don't think you need to cast the keys.
You can just say:
```
list(plot_values)[0]
```
In fact I think if you didn't wanna cast the whole thing to a list you could just say:
```
next(iter(plot_values))
```
`iter()` will turn the object into an iterator. `next()` exhausts the next item in the iterator. In this case the first one.

By default when you iterate over a dictionary, you get the keys back. It's kinda like looping over an array representing the keys. Hence why you don't need to explicitly call `keys()`.

But this is all trivia lol, what you've done is totally fine 😂

phill-stanley: Oh nice. Yeah I like the `next(iter(plot_values))` approach. It is cleaner, I've updated this.

A-Ashiq (approved):

PR #1913, opened 2024-11-04
Task/cdd 2277 updates to cms form page

phill-stanley:
# Description

This PR includes updates UKHSAPage model to include `last_published_date` and `body` field. As well as updates to several page models to remove `related_links` and `related_links_layout` including:

- landing page
- whats new
- metrics documentation

Fixes #CDD-2277

---

## Type of change

---

# Checklist:


A-Ashiq (approved):

cms/dashboard/models.py
```diff
@@ -24,6 +43,7 @@ class UKHSAPage(Page):
 
     """
 
+    body = RichTextField(features=AVAILABLE_RICH_TEXT_FEATURES)
```
A-Ashiq, line 46: Have you checked what the impact is here on all the sub-pages that extend from our base page? Is the frontend expecting certain things that we no longer provide and vice versa, can the CMS now provide things that the FE can't handle?

phill-stanley: Yes, there is some impact here with this ticket and the front-end, the work to align this is already in progress and will be tested against this branch.

The main impact to the front-end is from the removal of `related_links` and `related_links_layout` from some page types, which the front-end was treating as a required field even when unused or not part of the page's design.

The change with the body field should not have any impact here, each page returns a body field this update moves the basic body field type from `Common` to `UKHSAPage` and is overridden in types like `Composite` where additional features are required. So the body field returned by each page won't change in this update, we just provide it to any new page models created from `UKHSAPage`.

cms/composite/migrations/0012_alter_compositepage_body.py
```diff
@@ -0,0 +1,131 @@
+# Generated by Django 5.1.2 on 2024-10-25 14:15
```
A-Ashiq, line 1: So there is another open PR here -> https://github.com/UKHSA-Internal/data-dashboard-api/pull/1913 
which also adds migrations at the same nodes in the migration tree that you have so you're gonna need to co-ordinate between that PR. i.e. one them has to go in first and then the other will need to pull from main and then update the migration refs accordingly

PR #1918, opened 2024-11-05
Chore/cdd 2283 a

luketowell:
# Description

- Removal of the date_posted attribute from all models apart from the whats new parent and the whats new challenge pages
- Created all of the relevant migrations for the date_posted removal
- Updated all relevant tests

Fixes #CDD-<ISSUE_ID>

---

## Type of change

---

# Checklist:

- [x] I have deployed to the AWS environment and tested on machine


cms/common/migrations/0005_remove_commonpage_date_posted.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 5.1.2 on 2024-11-05 13:36
```
A-Ashiq, line 1: So there is another open PR here -> https://github.com/UKHSA-Internal/data-dashboard-api/pull/1913 
which also adds migrations at the same nodes in the migration tree that you have so you're gonna need to co-ordinate between that PR. i.e. one them has to go in first and then the other will need to pull from main and then update the migration refs accordingly

tests/unit/caching/test_internal_api_client.py
```diff
@@ -167,7 +167,7 @@ def test_build_headers(self, force_refresh: bool, cache_check_only: bool):
 
     @pytest.mark.parametrize(
         "additional_query_parameters",
-        ({"fields": "*"}, {"fields": "date_posted"}),
+        ({"fields": "*"}, {}),
```
A-Ashiq, line 170: Do we need to test the case of an empty dict i.e. no additional query params here?

cms/common/migrations/0005_remove_commonpage_date_posted.py
luketowell: I will reach out to Phill and see when he will be ready to merge. I also need to coordinate this PR with the FE PR https://github.com/UKHSA-Internal/data-dashboard-frontend/pull/539

tests/unit/caching/test_internal_api_client.py
luketowell: I am reading through the test cast and I think it might be worth keeping in the fields and the test for additional query params. I think the test is more for passing additional query params, and "date_posted" is an example of an additional param rather than related to date_posted. 

TLDR: Is it worth reverting this change as I think the test is valid and I got a little trigger happy with removing the contents of the dictionary?

A-Ashiq: Yeah I actually think it's better to keep it as it was, the date_posted is still valid for the whats new child page type

luketowell: I agree, I have reverted the change back to what it was previously 👍

A-Ashiq (approved):

PR #1956, opened 2024-11-20
CDD-2320: Provide smoothing of simplified line charts as option in cms

phill-stanley:
# Description

This PR includes an update to the CMS to provide a checkbox to enable/disable smoothing of `simplified line charts`:

Fixes #CDD-2320

---

## Type of change

---

# Checklist:


phill-stanley:
Screenshot 2024-11-20 at 10 12 31

cms/dynamic_content/elements.py
```diff
@@ -165,3 +165,8 @@ class SimplifiedChartPlotElement(BaseMetricsElement):
         required=False,
         help_text=help_texts.DATE_TO_FIELD,
     )
+    use_smooth_lines = blocks.BooleanBlock(
+        default=False,
```
A-Ashiq, line 169: Have you confirmed with DPD that it should default to using linear lines for these chart types?
For context,  we default to True for the other chart types for this option

phill-stanley: I'll confirm this with DPD, I'd defaulted to this because the ticket read like the smoothed lines were unlikely to be used (quote from the ticket below) but the AC does not include any suggestion of which should be the default value.

`The use of this "curved" line and appearance of smoothed data also goes against public health and epidemiological convention. Changing this would bring us in line with convention.`

A-Ashiq: Yeah let's double check with DPD before merging

phill-stanley: This has been approved, charts going forward will default to `linear` with the option to convert to smooth @A-Ashiq

A-Ashiq:
Won't you need a db migration for this?

phill-stanley:
> Won't you need a db migration for this? Well its kinda optional, i.e. it will work without one. But it will ask to make one whenever we try to migrate

Yeah, good shout @A-Ashiq I'd forgot to commit these after reseting my env while testing. I've pushed the migration files now.

PR #2002, opened 2024-12-10
Switch suggestions endpoint to submit form responses via AWS SES

A-Ashiq:
# Description

- Switches the suggestions v2 endpoint to send the form responses via AWS SES

Fixes #CDD-2364

---

## Type of change

---

# Checklist:


tests/integration/feedback/views/test_suggestions.py
```diff
@@ -68,10 +77,37 @@ def test_post_request_returns_correct_response(self):
...
+    @pytest.mark.django_db
+    def test_no_recipient_returns_bad_request(self, monkeypatch: MonkeyPatch):
+        """
+        Given a valid payload containing a question and answer suggestion
+        But no `FEEDBACK_EMAIL_RECIPIENT_ADDRESS` env var set
+        When the `POST /api/suggestions/v2/` endpoint is hit
+        Then an HTTP 400 BAD REQUEST response is returned
+        """
+        # Given
+        valid_payload = {
```
ChristianAMartin, line 97: I have only just joined the team so please disregard my suggestions if I have misunderstood the context but maybe it would be a good idea to use monkeypatch here to unset the env variable? e.g :-
`monkeypatch.delenv("FEEDBACK_EMAIL_RECIPIENT_ADDRESS", raising=False)`

tests/integration/feedback/views/test_suggestions.py
```diff
@@ -68,10 +77,37 @@ def test_post_request_returns_correct_response(self):
...
+
+        # When
+        response: Response = api_client.post(
+            path=self.path,
+            data=valid_payload,
+            format="json",
+        )
+
+        # Then
+        assert response.status_code == HTTPStatus.BAD_REQUEST.value
```
ChristianAMartin, line 113: Maybe we could add an error message to make the test failure more informative? e.g.
`assert response.status_code == HTTPStatus.BAD_REQUEST.value, "Expected HTTP 400 BAD REQUEST when FEEDBACK_EMAIL_RECIPIENT_ADDRESS is not set"`

tests/integration/feedback/views/test_suggestions.py
A-Ashiq: Ahh good catch, I'd started to do this cause I'd brought in the monkey patch fixture, and then forgot 🤦

tests/integration/feedback/views/test_suggestions.py
A-Ashiq: I've tended to shy away from this pattern as I'd expect that context to be in the docstring for the test. But happy to whack it in. I suppose this way the test runner would output that and know about it too

A-Ashiq: If this is something you wanna push, we might want to update the development standards doc at `docs/development_standards.md` and adopt it for tests being written moving forwards. Of course we will have thousands of tests in the existing suite which won't have it

ChristianAMartin: Hmm, in that case, let's keep it consistent with the rest and leave it as it is :)

ChristianAMartin (approved):

PR #2050, opened 2025-01-06
CDD-2371: Add support for manually setting min and max y-axis range

phill-stanley:
# Description

This PR adds support for manually providing a minimum and maximum value to the simplified line chart's y-axis range

Fixes #CDD-2371

---

## Type of change

---

# Checklist:


phill-stanley:
Screenshot 2025-01-06 at 09 44 44
Screenshot 2025-01-06 at 09 45 49

cms/dynamic_content/help_texts.py
```diff
@@ -220,6 +220,18 @@
 If nothing is provided, `metric value` will be used by default.
 """
 
+CHART_Y_AXIS_MINIMUM_VALUE: str = """
```
A-Ashiq, line 223: Good explanations

metrics/api/serializers/charts.py
```diff
@@ -102,6 +103,18 @@ class ChartsSerializer(serializers.Serializer):
         help_text=help_texts.CHART_Y_AXIS,
         default=DEFAULT_Y_AXIS,
     )
+    y_axis_minimum_value = serializers.IntegerField(
```
A-Ashiq, line 106: Will this always be an integer? I'm thinking of things like rates

phill-stanley: Good shout, in the case of simplified charts it would be but not if we extend this to other charts. I've updated the field type to `float`.

metrics/domain/charts/chart_settings.py
```diff
@@ -24,6 +30,14 @@ def width(self) -> int:
     def height(self) -> int:
         return self._chart_generation_payload.chart_height
 
+    @property
+    def y_axis_minimum_value(self) -> int | None:
```
A-Ashiq, line 34: Is this ever `None`? Don't we cast a default of 0 further upstream?

metrics/domain/charts/chart_settings.py
```diff
@@ -105,6 +119,53 @@ def get_line_with_shaded_section_chart_config(self):
...
+
+            if a `y_axis_maximum_value` is provided that is lower than the
+            highest value in the `y_axis_values` list then the max value of
+            `y_axis_values` is used as the highest number in the y-axis range.
+
+        Returns:
+            A dictionary containing the y-axis parameters than make up the
+            y-axis ticks and tick values.
+        """
+        plots_data = self.plots_data[0]
```
A-Ashiq, line 140: What happens if we have multiple plot where the later ones have larger Y values? If plot 1 has a max of say 10, but plot 2 goes up to 20. Then what happens there?
I think we need to consider every plot in the chart. Unless we're saying we only support this for the landing page single plot charts. Which I still don't know if that was agreed / holds true

phill-stanley: This method is used only to build out y_axis params for simplified charts, which only support a single plot.

I've included a ticket in backlog to add support for other chart types as a follow up. For the time being if you supply `y_axis_minimum_value` or `y_axis_maximum_value` while requesting any other chart type those attributes are just ignored.

metrics/domain/charts/chart_settings.py
phill-stanley: No, good catch. minimum will always default to 0. I'll update this.

PR #2061, opened 2025-01-09
Write static v2 charts to `tmp/` dir so that they are written into mounted volume

A-Ashiq:
# Description

- Write the v2 static charts (which is only used for debugging) to the `tmp/` dir. This is so that in the deployed envs the chart is written to a writable location. Note that this API already cleans up the file after. 

---

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

---

# Checklist:

### DIFF
--- a/metrics/interfaces/charts/access.py
+++ b/metrics/interfaces/charts/access.py
@@ -1,7 +1,7 @@
-import pathlib
 import urllib.parse
 from dataclasses import dataclass
 from datetime import datetime
+from pathlib import Path
 
 import plotly.graph_objects
 from django.db.models import Manager
@@ -540,7 +540,7 @@ def generate_chart_as_file(*, chart_request_params: ChartRequestParams) -> str:
     charts_interface = ChartsInterface(chart_request_params=chart_request_params)
     chart_output: ChartOutput = charts_interface.generate_chart_output()
 
-    pathlib.Path.mkdir("tmp", exist_ok=True)
+    Path("tmp").mkdir(exist_ok=True)
     return charts_interface.write_figure(figure=chart_output.figure)
 
 
--- a/metrics/interfaces/charts/access.py
+++ b/metrics/interfaces/charts/access.py
@@ -1,3 +1,4 @@
+import pathlib
 import urllib.parse
 from dataclasses import dataclass
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


metrics/interfaces/charts/access.py
```diff
@@ -539,6 +540,7 @@ def generate_chart_as_file(*, chart_request_params: ChartRequestParams) -> str:
     charts_interface = ChartsInterface(chart_request_params=chart_request_params)
     chart_output: ChartOutput = charts_interface.generate_chart_output()
 
+    pathlib.Path.mkdir("tmp", exist_ok=True)
```
A-Ashiq, line 543: Make the `tmp` directory before writing the chart. If used locally this will leave an empty `tmp` directory in place. If we think that's annoying I'll follow up and clean up that empty folder. I'd originally wanted to wrap into a context manager, something like:
```
with tempfile.TemporaryDirectory(prefix="tmp") as temp_dir:
```
But the nesting looked quite ugly with a try except block within it

metrics/interfaces/charts/access.py
```diff
@@ -471,7 +472,7 @@ def write_figure(self, *, figure: plotly.graph_objects.Figure) -> str:
             The filename of the image
 
         """
-        filename = f"new_chart.{self.chart_request_params.file_format}"
+        filename = f"tmp/new_chart.{self.chart_request_params.file_format}"
```
ChristianAMartin, line 475: I think we should check whether the new tmp directory exists before attempting to write the file e.g.
```
output_dir = Path("tmp")
output_dir.mkdir(exist_ok=True)

filename = output_dir / f"new_chart.{self.chart_request_params.file_format}"

try:
    figure.write_image(
        file=str(filename),
        format=self.chart_request_params.file_format,
        validate=False,
    )
except ValueError as e:
    raise RuntimeError(f"Failed to write image: {e}")

return str(filename.resolve())
```

phill-stanley (approved):

metrics/interfaces/charts/access.py
A-Ashiq: This happens upstream, but you're right it should move to the start of this method so its clear that it actually happens

A-Ashiq: That being said, I'm gonna hold off merging this in as I've gotta pick up something else with a higher priority. This also relates to a piece of infra that hasn't been merged in yet so it's nothing to worry about

PR #2067, opened 2025-01-10
Task/remove deprecated email logic/cdd 2364

A-Ashiq:
# Description

- Removes the env vars associated with deprecated email approach
- Removes versioned suffixes from function names so it's clear there is only the 1 approach now

Fixes #CDD-2364

---

## Type of change

---

# Checklist:


feedback/email_server.py
```diff
@@ -121,7 +85,7 @@ def send_email_v3(
         send_email_via_ses(email_message=email_message)
```
ChristianAMartin, line 85: why do we want to suppress any send errors?
```suggestion
    try:
        send_email_via_ses(email_message=email_message)
    except ClientError:
        logger.error("Failed to send email via SES.")
```

A-Ashiq: We already log out within the function, no need for us to record another log which contains less information than the 1st

phill-stanley (approved):

PR #2180, opened 2025-02-12
CDD-2432 Implement Middleware - Adds permission models

mxrman:
# Description

-

Fixes #CDD-<ISSUE_ID>

---

## Type of change

---

# Checklist:


metrics/data/models/api_models/api_groups.py
```diff
@@ -0,0 +1,16 @@
+from django.db import models
+
+
+class ApiGroup(models.Model):
+
+    class Meta:
+        db_table = "api_groups"
```
A-Ashiq, line 7: I think we need to be very careful about how we use the term `api` here. In the context of the data layer, `api` generally refers to the denormalized flat table which serves the public API service.

metrics/data/models/api_models/api_permissions.py
```diff
@@ -0,0 +1,141 @@
...
+    def clean(self):
+        if self.sub_theme and self.theme and not self.sub_theme.theme == self.theme:
+            raise AdminFormSubthemeAssocThemeError
+        if self.sub_theme and self.topic:
+            if not self.topic.sub_theme.theme == self.theme:
+                raise AdminFormTopicAssocSubthemeError
+            if not self.sub_theme.name == self.topic.sub_theme.name:
+                raise AdminFormSubthemeAssocTopicError
+
+        api_permission = ApiPermission.objects.filter(
```
A-Ashiq, line 126: Generally we don't want to build and reference queries like this directly.
We should probs move this to the corresponding model manager and abstract it away from the reader here

metrics/data/models/api_models/api_permissions.py
```diff
@@ -0,0 +1,141 @@
...
+            model_str += f", age={self.geography.name}"
+        if self.stratum is not None:
+            model_str += f", stratum={self.stratum.name}"
+        return model_str
+
+    def clean(self):
+        if self.sub_theme and self.theme and not self.sub_theme.theme == self.theme:
+            raise AdminFormSubthemeAssocThemeError
+        if self.sub_theme and self.topic:
+            if not self.topic.sub_theme.theme == self.theme:
```
A-Ashiq, line 121: I have a feeling there may be some linting tools that would be unhappy about this. Possibly sonar won't like the not `<condition> ==`, it will probably tell you to use `!=` instead

metrics/data/models/api_models/api_permissions.py
```diff
@@ -0,0 +1,141 @@
...
+            model_str += f", geography_type={self.geography_type.name}"
+        if self.geography is not None:
+            model_str += f", geography={self.geography.name}"
+        if self.age is not None:
+            model_str += f", age={self.geography.name}"
+        if self.stratum is not None:
+            model_str += f", stratum={self.stratum.name}"
+        return model_str
+
+    def clean(self):
```
A-Ashiq, line 117: I know the super() clean is empty but it's worth calling it here just in case something does get added there that we'd otherwise be overriding

metrics/data/models/api_models/api_permissions.py
```diff
@@ -0,0 +1,141 @@
...
+        if self.sub_theme and self.theme and not self.sub_theme.theme == self.theme:
```
A-Ashiq, line 118: Would we ever have a situation where `self.theme` has not been populated?

metrics/data/models/api_models/api_permissions.py
```diff
@@ -0,0 +1,141 @@
...
+            raise AdminFormSubthemeAssocThemeError
+        if self.sub_theme and self.topic:
+            if not self.topic.sub_theme.theme == self.theme:
+                raise AdminFormTopicAssocSubthemeError
+            if not self.sub_theme.name == self.topic.sub_theme.name:
```
A-Ashiq, line 123: I think we'd be safer to compare the objects not the names as you've done above because of this thing of having multiple entries with the same name being a valid possibility

tests/integration/metrics/data/models/api_models/test_api_group.py
```diff
@@ -0,0 +1,63 @@
...
+        self.permission2 = ApiPermission.objects.create(
+            name="permission2",
+            theme=self.theme,
+            sub_theme=self.sub_theme,
+            topic=self.topic2,
+            metric=self.metric2,
+        )
+
+    @pytest.mark.django_db
+    def test_create_api_group(self):
```
A-Ashiq, line 41: Can we please follow the documented test structure for these

tests/integration/metrics/data/models/api_models/test_api_group.py
```diff
@@ -0,0 +1,63 @@
...
+    def test_api_group_str(self):
+        assert str(self.api_group) == "admin_group"
+
+    @pytest.mark.django_db
+    def test_add_permissions_to_api_group(self):
+        self.api_group.permissions.add(self.permission1, self.permission2)
+        assert self.api_group.permissions.count() == 2
+
+    @pytest.mark.django_db
+    def test_remove_permission_from_api_group(self):
```
A-Ashiq, line 55: Can we please follow the documented test structure for these tests

metrics/data/models/api_models/api_groups.py
```diff
@@ -0,0 +1,16 @@
+from django.db import models
```
A-Ashiq, line 1: I don't think we should put these models here for the reason below. In the context of the data layer, `api_models` refers to the denormalized flat table which serves the public API

metrics/data/models/api_models/api_groups.py
mxrman: How about `metric_auth_permissions` &`metric_auth_groups`?

metrics/data/models/api_models/api_permissions.py
mxrman: I actually need to check here for `self.theme` & `self.sub_theme` or else, instead of the form validation exceptions a `RelatedObjectDoesNotExist` is raised. e.g:
```
        if not getattr(self, "theme", None) or not getattr(self, "sub_theme", None):
            raise AdminFormThemeSubthemeError
```
fixes this issue.

mxrman:
Moved to here -> https://github.com/UKHSA-Internal/data-dashboard-api/pull/2195

PR #2195, opened 2025-02-17
CDD-2432 Implement Middleware - Adds permission models

mxrman:
# Description

Adds The following models to the data app.
- `api_groups`
Screenshot 2025-02-12 at 15 41 49

- `api_groups_permissions`
Screenshot 2025-02-12 at 15 42 56

- `api_permissions`

Screenshot 2025-02-12 at 15 43 52

Also adds some organising of the existing api models: Moves the `APITimeSeriesFactory` model into a `api_models` package (including tests).

Fixes #CDD-2432

---

## Type of change

---

# Checklist:


metrics/data/models/rbac_models/rbac_permissions.py
```diff
@@ -0,0 +1,147 @@
...
+            model_str += f", geography={self.geography.name}"
+        if self.age is not None:
+            model_str += f", age={self.age.name}"
+        if self.stratum is not None:
+            model_str += f", stratum={self.stratum.name}"
+        return model_str
+
+    def clean(self):
+        super().clean()
+        if not getattr(self, "theme", None) or not getattr(self, "sub_theme", None):
```
A-Ashiq, line 130: Is there a reason we need to use `getattr`, the reader might look at this and think that the `self.theme` attribute itself is not always gonna be there?
But really we just wanna check for falsiness, so wouldn't it be okay to just call self.theme and self.sub_theme directly? If they haven't been populated then they will be None anyway

metrics/data/models/rbac_models/rbac_permissions.py
```diff
@@ -0,0 +1,147 @@
...
+            raise AdminFormThemeSubthemeError
+        if self.sub_theme.theme != self.theme:
+            raise AdminFormSubthemeAssocThemeError
+        if self.topic:
```
A-Ashiq, line 134: This method feels like it could be broken up a little further, maybe we can wrap each of these blocks into their own method and just call out to them here instead?
```
super().clean()
self._validate_theme_selection
self._validate_sub_theme_selection
self._validate_topic_selection

# check for duplicate permissions
```

tests/factories/metrics/rbac_models/rbac_permission.py
```diff
@@ -0,0 +1,96 @@
...
+    GeographyType,
+    Metric,
+    Stratum,
+    SubTheme,
+    Theme,
+    Topic,
+)
+
+
+class ApiPermissionFactory(factory.django.DjangoModelFactory):
```
A-Ashiq, line 18: Should this be called `RBACPermissionFactory`?

tests/factories/metrics/rbac_models/rbac_permission.py
```diff
@@ -0,0 +1,96 @@
...
+
+    class Meta:
+        model = RBACPermission
+
+    @classmethod
+    def create_record(
+        cls,
+        name: str = "permission1",
```
A-Ashiq, line 26: Nitpick: I'd say let's avoid suffixing with a number even on factory defaults, maybe give it something like `all_infectious_respiratory_data`

tests/integration/metrics/data/managers/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,91 @@
+import pytest
+from metrics.data.models.rbac_models import (
+    RBACPermission,
+    AdminFormDuplicatePermissionError,
+)
+from tests.factories.metrics.rbac_models.rbac_permission import ApiPermissionFactory
+
+
+class TestApiPermissionFactory:
```
A-Ashiq, line 9: Same comment as before, this needs to be renamed to `TestRBACPermissionFactory`

tests/integration/metrics/data/managers/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @pytest.mark.django_db
+    def test_create_record_creates_valid_permission(self):
+        """
+        Given valid input parameters,
+        When `create_record` is called,
+        Then an `RBACPermission` instance is created with the correct associations.
+        """
+        # When creating a permission record
```
A-Ashiq, line 29: Nitpick: we don't need to repeat this here. We can put this info in the test docstring:
```
"""
Given a valid set of input parameters
When `create_record` is called from the `RBACPermissionFactory`
Then an `RBACPermission` instance is created with the correct associations
"""
# Given
the block to arrange the state

# When
do the action

# Then
do the assertions

```
The #Given, #When, #Then should just be single word comments at the top of each block signifiying the start of that clause.

tests/integration/metrics/data/managers/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,91 @@
...
+        And a unique `RBACPermission` record
+        When `get_existing_permissions()` is called
+        Then only the records with matching fields (excluding the given instance) are returned
+        """
+        # Given creating an initial permission
+        permission = ApiPermissionFactory.create_record()
+
+        # Create another permission with different `name`, `theme` & `sub_theme` (to ensure uniqueness)
+        _ = ApiPermissionFactory.create_record(
+            name=f"permission_2",
```
A-Ashiq, line 58: You can remove the `f` string here

metrics/data/managers/rbac_models/rbac_permissions.py
```diff
@@ -0,0 +1,39 @@
+from django.db import models
+
+
+class RBACPermissionQuerySet(models.QuerySet):
+    """Custom queryset which can be used by the `RBACPermissionManager`."""
+
+    def get_existing_permissions(
```
A-Ashiq, line 7: Would we ever expect to see multiple existing permissions?

tests/integration/metrics/data/managers/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,91 @@
...
+        assert retrieved_permissions.count() == 0
+
+    @pytest.mark.django_db
+    def test_create_duplicate_permission_raises_error(self):
+        """
+        Given an existing `RBACPermission` record with specific fields,
+        When another permission with the same fields is created,
+        Then an `AdminFormDuplicatePermissionError` should be raised.
+        """
+        # Given Create an initial permission
```
A-Ashiq, line 77: Same comment as above regarding our adopted test structure

tests/integration/metrics/data/models/rbac_models/test_rbac_group_permissions.py
```diff
@@ -0,0 +1,103 @@
...
+        self.topic = Topic.objects.create(name="asthma", sub_theme=self.sub_theme)
+        self.metric1 = Metric.objects.create(
+            name="COVID-19_vaccinations_autumn22_dosesByDay"
+        )
+        self.metric2 = Metric.objects.create(name="COVID-19_deaths_ONSByWeek")
+
+        self.geography_type = GeographyType.objects.create(name="Nation")
+        self.geography = Geography.objects.create(name="England")
+        self.age = Age.objects.create(name="all")
+        self.stratum = Stratum.objects.create(name="default")
```
A-Ashiq, line 30: If you wanted to, you could just use the `CoreTimeSeriesFactory` which creates all these for you. See `tests/factories/metrics/time_series.py`

tests/integration/metrics/data/models/rbac_models/test_rbac_group_permissions.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        Given an existing RBACGroupPermission
+        When another group with the same name is attempted to be created
+        Then a database IntegrityError should be raised
+        """
+        # Given
+        RBACGroupPermission.objects.create(name="duplicate_group")
+
+        # When / Then
+        with pytest.raises(Exception):  # Adjust exception type based on DB backend
```
A-Ashiq, line 84: We shouldn't leave the broad exception here. This one should do the trick:
```
django.db.utils.IntegrityError 
```

tests/integration/metrics/data/models/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        Given an RBACPermission without a theme or subtheme
+        When it is validated
+        Then an AdminFormThemeSubthemeError is raised
+        """
+        # Given / When / Then
+        with pytest.raises(AdminFormThemeSubthemeError):
+            RBACPermission(name="invalid_permission").clean()
+
+        with pytest.raises(AdminFormThemeSubthemeError):
```
A-Ashiq, line 85: This feels like 2 tests, can we break them up?

tests/integration/metrics/data/models/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,187 @@
...
+        When an RBACPermission is created referencing this subtheme
+        Then an AdminFormSubthemeAssocTopicError is raised
+        """
+        # Given
+        sub_theme_without_topic = SubTheme.objects.create(
+            name="mpox-clade-1b", theme=self.non_communicable
+        )
+
+        # When / Then
+        with pytest.raises(AdminFormSubthemeAssocTopicError):
```
A-Ashiq, line 147: Does this mean that we can't wildcard for a subtheme?
For example, as an admin user I want to give group XYZ the permission to see all of `theme=infectious_disease, sub_theme=respiratory`?
This says that we have the provide a topic as well?

metrics/data/models/rbac_models/rbac_permissions.py
mxrman: If this foreign key is not nullable, Django raises a RelatedObjectDoesNotExist exception when trying to access a related object that is None... e.g. 

The other option is to let Theme and SubTheme foreign keys to be nullable but then we are just relying on validation in the clean method?
```
  File "/Users/joegasewicz/Work/data-dashboard-api-2/metrics/data/models/rbac_models/rbac_permissions.py", line 130, in clean
    if not self.theme or not self.sub_theme:
           ^^^^^^^^^^
  File "/Users/joegasewicz/Work/data-dashboard-api-2/.venv/lib/python3.12/site-packages/django/db/models/fields/related_descriptors.py", line 267, in __get__
    raise self.RelatedObjectDoesNotExist(
metrics.data.models.rbac_models.rbac_permissions.RBACPermission.theme.RelatedObjectDoesNotExist: RBACPermission has no theme.
```

tests/integration/metrics/data/models/rbac_models/test_rbac_permissions.py
A-Ashiq: In that case I think we need to re-word the docs and the name of this test slightly to suit

metrics/data/managers/rbac_models/rbac_permissions.py
A-Ashiq: The unique constraint would presumably prevent any race condition? What kind of race condition were you thinking of here?

metrics/data/models/rbac_models/rbac_permissions.py
A-Ashiq: The docstrings feel like the better approach here, thanks 👍

tests/integration/metrics/data/models/rbac_models/test_rbac_group_permissions.py
```diff
@@ -0,0 +1,103 @@
...
+    @pytest.mark.django_db
+    def test_rbac_group_can_have_multiple_permissions(self):
+        """
+        Given an `RBACGroupPermission`
+        When multiple unique `RBACPermission` instances are added
+        Then the group successfully stores and retrieves all permissions
+        """
+        # Given
+        rbac_group = RBACGroupPermission.objects.create(name="respiratory_admins")
+        rbac_group.permissions.add(
```
A-Ashiq, line 96: This is the When clause:
```
# Given
rbac_group = RBACGroupPermission.objects.create(name="respiratory_admins")

# When
rbac_group.permissions.add(self.permission_dose_by_day, self.permission_ons_by_week)

# Then
...
```

tests/integration/metrics/data/models/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,216 @@
...
+    GeographyType,
+    Geography,
+    Age,
+    Stratum,
+)
+
+
+class TestRBACPermission:
+    @pytest.fixture(autouse=True)
+    def setup(self, db):
```
A-Ashiq, line 25:
```suggestion
    def setup(self):
```

metrics/data/managers/rbac_models/rbac_permissions.py
mxrman: If multiple transactions try to add the same permission's record at the same time, there might be a brief moment where multiple identical permissions exist before one of them gets rolled back due to the unique constraint.
But it's probably highly unlikely with the amount of authorised users we have accessing the private api?

metrics/data/models/rbac_models/rbac_permissions.py
A-Ashiq: I'm thinking it might have been clearer to try except instead, that way it's a bit more obvious what's going on

metrics/data/managers/rbac_models/rbac_permissions.py
A-Ashiq: The `RBACPermission` gets created by the admin console user. Why would users be creating permissions when hitting the private API?

A-Ashiq: Am I misunderstanding how we are intending to use this model?

mxrman: No, you're correct, by user I meant admin console user.

A-Ashiq: So why would we be creating those records when hitting the private api? Surely that's essentially a read-only operation?

mxrman: I meant, if 2 admin console users both created the same permissions in the Django admin at the same time there would be a very slim chance of a race condition

tests/integration/metrics/data/models/rbac_models/test_rbac_group_permissions.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        Given an existing RBACGroupPermission
+        When another group with the same name is attempted to be created
+        Then a database IntegrityError should be raised
+        """
+        # Given
+        RBACGroupPermission.objects.create(name="duplicate_group")
+
+        # When / Then
+        with pytest.raises(IntegrityError):  # Adjust exception type based on DB backend
```
A-Ashiq, line 86: This comment is no longer relevant, can you remove this please

tests/integration/metrics/data/models/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,216 @@
...
+        Then an AdminFormSubthemeAssocThemeError is raised
+        """
+        # Given / When / Then
+        with pytest.raises(AdminFormSubthemeAssocThemeError):
+            RBACPermission(
+                name="invalid_permission", theme=self.non_communicable
+            ).clean()
+
+    @pytest.mark.django_db
+    def test_subtheme(self):
```
A-Ashiq, line 99: We can't have duplicate names of tests. This has the same name as above

tests/integration/metrics/data/models/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,216 @@
...
+                name="invalid_topic_permission",
+                theme=self.non_communicable,
+                sub_theme=self.respiratory,
+                topic=topic_with_different_theme,
+            ).clean()
+
+    @pytest.mark.django_db
+    def test_rbac_permission_rejects_topic_unrelated_to_subtheme(self):
+        """
+        GIVEN a Topic that belongs to a different SubTheme
```
A-Ashiq, line 166: Nitpick: Don't capitilze the Given, When, Then clauses

tests/integration/metrics/data/models/rbac_models/test_rbac_group_permissions.py
```diff
@@ -0,0 +1,107 @@
...
+    Theme,
+    SubTheme,
+    Topic,
+)
+
+
+class TestRBACGroupPermission:
+
+    @pytest.fixture(autouse=True)
+    def setup(self, db):
```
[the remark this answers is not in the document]
mxrman: Sorry, me bad, I removed another arg from a different file

metrics/data/models/rbac_models/rbac_permissions.py
```diff
@@ -0,0 +1,162 @@
...
+            model_str += f", stratum={self.stratum.name}"
+        return model_str
+
+    def clean(self):
+        super().clean()
+        self._validate_theme_selection()
+        self._validate_subtheme_selection()
+        self._validate_topic_selection()
+
+        rbac_permissions = self._get_existing_permissions()
```
A-Ashiq, line 135: I'd previosly left a comment on this and it looks like its been resolved without being actioned again. Can you please wrap this in a private method and call out to it like we've done above

metrics/data/models/rbac_models/rbac_permissions.py
A-Ashiq: @mxrman what do you think of wrapping it in a try except around the `RelatedObjectDoesNotExist` error so it's clear about what is happening, instead of swallowing it in the getattr call?

tests/factories/metrics/rbac_models/rbac_permission.py
```diff
@@ -0,0 +1,96 @@
...
+    class Meta:
+        model = RBACPermission
+
+    @classmethod
+    def create_record(
+        cls,
+        name: str = "all_infectious_respiratory_data",
+        theme_name: str = "infectious_disease",
+        sub_theme_name: str = "respiratory",
+        topic_name: str = None,
```
A-Ashiq, line 29: Nitpick: Arguably these type hints should be `str | None` because we pass `None` as the default arg

tests/integration/metrics/data/managers/rbac_models/test_rbac_permissions.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        Given multiple `RBACPermission` records with overlapping fields
+        And a unique `RBACPermission` record
+        When `get_existing_permissions()` is called
+        Then only the records with matching fields (excluding the given instance) are returned
+        """
+        # Given
+        permission = RBACPermissionFactory.create_record()
+
+        _ = RBACPermissionFactory.create_record(
```
A-Ashiq, line 57: Nitpick: Do you need to assign the throwaway variable here? Can you just call that method from the factory outright without assigning it?

tests/factories/metrics/rbac_models/rbac_permission.py
```diff
@@ -0,0 +1,96 @@
...
+        topic = None
+        if topic_name:
+            topic = Topic.objects.filter(name=topic_name, sub_theme=sub_theme).first()
+            if not topic and topic_name:
+                topic, _ = Topic.objects.get_or_create(
+                    name=topic_name, sub_theme=sub_theme
+                )
+
+        metric = None
+        if metric_name:
```
A-Ashiq, line 54: Out of curiosity, why not just do the get or create instead of i

tests/factories/metrics/rbac_models/rbac_group_permissions.py
```diff
@@ -0,0 +1,49 @@
...
+class RBACPermissionGroupFactory(factory.django.DjangoModelFactory):
+
+    class Meta:
+        model = RBACGroupPermission
+
+    @classmethod
+    def create_record(
+        cls,
+        name: str = "default",
+        permissions: list[RBACPermission] = [],
```
A-Ashiq, line 23: Oh actually, this is a mutable argument.
I think we should actually specify the default as None. Something like this:
```
permissions: list[RBACPermission] | None = None,
...

permissions = permissions or []
for permission in permission:
    ...
```

https://docs.python-guide.org/writing/gotchas/

A-Ashiq (approved):

PR #2216, opened 2025-02-25
Task/auth middleware cdd 2432 api view decorator

mxrman:
# Description

- Implemented `require_authorisation` decorator for enforcing RBAC-based authentication using the `X-GroupId` header, controlled by the `AUTH_ENABLED` setting declared in https://github.com/UKHSA-Internal/data-dashboard-api/blob/c481ab597641b865ef15bfe412470c32908c47f6/metrics/api/settings/private_api.py#L5.
- Added `_set_rbac_group_permissions` to retrieve and attach permissions to incoming requests.
- Created `RBACGroupPermissionQuerySet` & `RBACGroupPermissionManager` for permission retrieval.
- Added `MockDownloadView` and test cases to validate & simulate authorisation behavior.
- Added unit tests for `RBACGroupPermission` creation and duplicate record handling.

⚠️ This PR does not apply `require_authorisation` to any actual API views yet.

Fixes #CDD-2432

---

## Type of change

---

# Checklist:


metrics/data/managers/rbac_models/rbac_group_permissions.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def get_queryset(self) -> RBACGroupPermissionQuerySet:
+        """
+        Returns the custom queryset for RBACGroupPermission.
+
+        This allows access to custom queryset methods.
+        """
+        return RBACGroupPermissionQuerySet(self.model, using=self._db)
+
+    def get_group(self, name: str) -> "RBACGroupPermission":
```
A-Ashiq, line 33:
```suggestion
    def get_group(self, name: str) -> "RBACGroupPermission" | None:
```

metrics/api/decorators/auth.py
```diff
@@ -0,0 +1,26 @@
...
+
+RBAC_AUTH_X_HEADER = "X-GroupId"
+
+
+def require_authorisation(func):
+    @wraps(func)
+    def wrap(self, request, *args, **kwargs):
+        if not AUTH_ENABLED:
+            return func(self, request, *args, **kwargs)
+        if RBAC_AUTH_X_HEADER in request.headers:
```
A-Ashiq, line 14: Nitpick: Instead of the double `if` statement we could try and except instead?
```
try:
    group_id: str = request.headers[RBAC_AUTH_X_HEADER]
except KeyError:
    pass
else:
    _set_rbac_group_permissions(request=request, group_id=group_id)
```

metrics/api/decorators/auth.py
```diff
@@ -0,0 +1,26 @@
...
+            group_id = request.headers.get(RBAC_AUTH_X_HEADER)
+            if group_id:
+                _set_rbac_group_permissions(request, group_id)
+        return func(self, request, *args, **kwargs)
+
+    return wrap
+
+
+def _set_rbac_group_permissions(request, group_id: str) -> None:
```
A-Ashiq, line 23: Can we enforce use of key word arguments here?
```
def _set_rbac_group_permissions(*, request, group_id: str)
```
The `request` arg will also need a type hint

tests/integration/metrics/data/managers/rbac_models/test_rbac_group_permissions.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+
+from metrics.data.models.rbac_models import RBACGroupPermission
+from tests.factories.metrics.rbac_models.rbac_group_permissions import (
+    RBACPermissionGroupFactory,
+)
+
+
+class TestRBACGroupPermissionFactory:
```
A-Ashiq, line 9: Factories are just a tool that we leverage when testing our source code so it feels odd that we're writing tests to check how they work

tests/integration/metrics/api/decorators/test_auth.py
```diff
@@ -0,0 +1,115 @@
...
+from tests.factories.metrics.rbac_models.rbac_group_permissions import (
+    RBACPermissionGroupFactory,
+)
+from tests.factories.metrics.rbac_models.rbac_permission import RBACPermissionFactory
+
+
+MODULE_PATH = "metrics.api.decorators.auth"
+
+
+class MockApiView(APIView):
```
A-Ashiq, line 21:
```suggestion
class FakeApiView(APIView):
```
Nitpick: I know terminology varies a lot but we tend to distinguish between flexible mock objects like the ones you'd get from the `unittest` library and fakes. In this case, we'd say this is a fake because its a barebone implementation of the real thing.

tests/integration/metrics/api/decorators/test_auth.py
```diff
@@ -0,0 +1,115 @@
...
+    @require_authorisation
+    def post(self, request, *args, **kwargs):
+        permissions = getattr(request, "group_permissions", None)
```
A-Ashiq, line 24: Nitpick: you could simplifiy this like so:
```
try:
    permissions = request.group_permissions
except AttributeError:
    permissions = []

permissions_data = [p.name for p in permissions]

...
```
This is slightly simpler because you don't have to then treat the negative case of None

tests/integration/metrics/api/decorators/test_auth.py
```diff
@@ -0,0 +1,115 @@
...
+    def test_request_succeeds_when_auth_is_disabled(self):
+        """
+        Given authentication is disabled
+        When a request is made to an authorised route
+        Then the response is successful
+        """
+        # Given
+        client = APIClient()
+
+        with mock.patch(f"{MODULE_PATH}.AUTH_ENABLED", False):
```
A-Ashiq, line 53: Nitpick: Why not apply this as a decorator to the test as well? The rest of the test suite tends to prefer that approach unless you have to enforce that condition for a subset of the test only

tests/integration/metrics/api/decorators/test_auth.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Given authentication is enabled
+        And a valid `X-GroupId` header is provided
+        When a request is made to an authorised route
+        Then the response is successful
+        """
+        # Given
+        client = APIClient()
+        headers = {f"HTTP_{RBAC_AUTH_X_HEADER}": "medical"}
+        all_infectious = RBACPermissionFactory.create_record(
```
A-Ashiq, line 73: Wouldn't this be `all_respiratory_data` because we've gone down a level and specified the sub theme as respiratory?

metrics/api/decorators/auth.py
```diff
@@ -0,0 +1,26 @@
...
+    group_permissions = RBACGroupPermission.objects.get_group(name=group_id)
+    if group_permissions:
+        request.group_permissions = list(group_permissions.permissions.all())
```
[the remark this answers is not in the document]
mxrman: No, currently there is no logic that expects a `group_permissions` on the `request` object.
There will be some work in #CDD-2520 that will use the  `group_permissions` attribute if it is available within a serialiser.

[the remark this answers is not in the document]
A-Ashiq: Does it make sense to always set that attribute and if we have no permissions just set it to an empty list?
That way the object is safe to interact with

metrics/api/decorators/auth.py
mxrman: This works for all conditions except if `group_id` is and empty string...

mxrman: yes that's the current behaviour but this is if block stops making an unnecessary database call with an empty string. If downstream there will never be a case of an empty string, either from an internal of external client then we could skip this check?

A-Ashiq: Perhaps then in `_set_rbac_group_permissions()` we set `group_permissions` on the request object to be an empty list if the group ID is falsy and skip having to fetch the rbac group permissions?

mxrman: Sounds great, I'll get that change in, thanks.

mxrman: Updated & added an extra test for that clause.

A-Ashiq (approved):

PR #2246, opened 2025-03-06
Task/cdd 2505 - Multiple Global Banners

luketowell:
# Description

- Removal of the constraint where only 1 Global Banner is available
- Return multiple banners to the front end if they have been marked as active

Fixes #CDD-2505

---

## Type of change

---

# Checklist:


cms/snippets/serializers/global_banner.py
```diff
@@ -31,7 +31,7 @@ def data(self) -> dict[str, ReturnDict[str, str] | None]:
         active_global_banner_data: ReturnDict[str, str] | None = (
             get_active_global_banner(global_banner_manager=self.global_banner_manager)
         )
-        return {"active_global_banner": active_global_banner_data}
+        return {"active_global_banners": active_global_banner_data}
```
A-Ashiq, line 34: This is introducing a breaking change to an API. We need to bump the global banners endpoint version.
Otherwise soon as you deploy this, the FE will break

tests/fakes/managers/cms/global_banner_manager.py
```diff
@@ -12,12 +12,12 @@ def __init__(self, global_banners: list[GlobalBanner], **kwargs):
...
+    def get_active_banners(self) -> GlobalBanner | None:
         try:
-            return next(
+            return [
                 global_banner
                 for global_banner in self.global_banners
                 if global_banner.is_active is True
-            )
+            ]
         except StopIteration:
```
A-Ashiq, line 22: You can remove the catch for the stop iteration error now because it will never get thrown. `StopIteration` gets thrown when an iterator is exhausted i.e. when we call `next()` on something and there's nothing left to give

cms/snippets/managers/global_banner.py
```diff
@@ -33,7 +33,7 @@ def has_active_banner(self) -> bool:
         """
         return self.get_queryset().get_active_banners().exists()
 
-    def get_active_banner(self):
+    def get_active_banners(self):
         """Gets the currently active `GlobalBanner`.
```
A-Ashiq, line 37:
```suggestion
        """Gets the currently active `GlobalBanner` objects.
```

cms/snippets/managers/global_banner.py
```diff
@@ -42,4 +42,4 @@ def get_active_banner(self):
             then None is returned.
```
A-Ashiq, line 42: Can you amend this docstring so it doesn't reference the empty case as None

tests/fakes/managers/cms/global_banner_manager.py
```diff
@@ -12,12 +12,12 @@ def __init__(self, global_banners: list[GlobalBanner], **kwargs):
         self.global_banners = global_banners
         super().__init__(**kwargs)
 
-    def get_active_banner(self) -> GlobalBanner | None:
+    def get_active_banners(self) -> GlobalBanner | None:
```
A-Ashiq, line 15:
```suggestion
    def get_active_banners(self) -> list[GlobalBanner]:
```

tests/unit/cms/snippets/serializers/test_global_banner.py
```diff
@@ -78,11 +78,71 @@ def test_serialized_data_for_active_global_banner(self):
...
+                {
+                    "title": active_global_banner.title,
+                    "body": active_global_banner.body,
+                    "banner_type": active_global_banner.banner_type,
+                }
+            ]
+        }
+        assert serializer.data == expected_data
+
+    def test_serialized_data_for_multiple_active_global_banner(self):
```
A-Ashiq, line 91: This test is nice 👌

cms/snippets/managers/global_banner.py
```diff
@@ -23,7 +23,7 @@ class GlobalBannerManager(models.Manager):
     def get_queryset(self) -> GlobalBannerQuerySet:
         return GlobalBannerQuerySet(model=self.model, using=self.db)
 
-    def has_active_banner(self) -> bool:
+    def has_active_banners(self) -> bool:
```
A-Ashiq, line 26: At this point. Do we have a need for this? Presumably we now don't need to check for if we have any active banners. We can always just plough ahead and make new active banners regardless?

A-Ashiq: In fact is there a need for `get_active_banners()` either?

cms/snippets/serializers/global_banner.py
luketowell: Yea, thats true. I wasn't planning on merging until I had updated the front end. Is endpoint versioning something that has been done elsewhere in the code base? If so can you point me in the direction of an example so I can follow the same standard?

A-Ashiq: We tend to remove old endpoints from the private API and just keep the ones in use. Since the FE is the only consumer of the private API, it doesn't make sense to keep them around. The main reason we do versioning in the private API is so we have backwards compatilibity for that transition period in which the FE is still pointing to an old version when a new one is available in the BE.
Let's jump on a call and I can talk you through how to do this

cms/snippets/serializers/global_banner.py
```diff
@@ -48,6 +63,26 @@ def get_active_global_banner(
         If no global banner is active, then None is returned
 
     """
-    global_banner = global_banner_manager.get_active_banner()
-    serializer = GlobalBannerResponseSerializer(instance=global_banner)
+    global_banners = global_banner_manager.get_active_banners().first()
```
A-Ashiq, line 66: The reason we use model managers is so that we can push this kind of stuff over into the model manager abstraction. So we don't leak that implementation detail into the caller.
I don't mind if you leave this one as you have it here, cause we'll be deleting this soon anyway, but just something to bear in mind

cms/snippets/managers/global_banner.py
A-Ashiq, line 26: At this point, do we need this?

tests/fakes/managers/cms/global_banner_manager.py
A-Ashiq, line 15: For `get_active_banners` we are probably gonna return `-> list[GlobalBanner]`

tests/fakes/managers/cms/global_banner_manager.py
A-Ashiq, line 22: We would no longer expect this `StopIteration` error to ever be thrown. In the case of no global banners, we would just return an empty list. 
The `StopIteration` error gets thrown when we exhaust an iterable, which we would have done with the `next()` keyword

tests/integration/cms/snippets/views/test_global_banner.py
```diff
@@ -11,13 +11,13 @@
 class TestGlobalBannerView:
     @property
     def path(self) -> str:
-        return "/api/global-banners/v1"
+        return "/api/global-banners/v2"
```
A-Ashiq, line 14: Same comment as above. We need to maintain the test coverage we had before on the deprecated endpoint for the time being.

cms/snippets/serializers/global_banner.py
```diff
@@ -48,6 +63,26 @@ def get_active_global_banner(
...
+    serializer = GlobalBannerResponseSerializer(instance=global_banners)
+    return serializer.data
+
+
+def get_active_global_banners(
+    global_banner_manager: Manager,
+) -> ReturnDict[str, str] | None:
```
A-Ashiq, line 73: Do we ever return None?

cms/snippets/serializers/global_banner.py
```diff
@@ -48,6 +63,26 @@ def get_active_global_banner(
...
+    """Gets the currently active global banner information
```
A-Ashiq, line 74:
```suggestion
    """Gets information about any currently active global banners
```

cms/snippets/serializers/global_banner.py
```diff
@@ -48,6 +63,26 @@ def get_active_global_banner(
...
+
+    Args:
+        `global_banner_manager`: The `GlobalBanner`
+            model manager used to query for records
+
+    Returns:
+        Dict representation the of the active global banner.
```
A-Ashiq, line 81:
```suggestion
        Dict representation of any active global banners.
```

cms/snippets/serializers/global_banner.py
```diff
@@ -48,6 +63,26 @@ def get_active_global_banner(
...
+        If no global banner is active, then None is returned
```
A-Ashiq, line 82: Should this be an empty list not None?

tests/integration/cms/snippets/managers/test_global_banner.py
```diff
@@ -18,7 +18,7 @@ def test_has_active_banner(self):
         GlobalBannerFactory.create(is_active=True)
 
         # When
-        has_active_banner: bool = GlobalBanner.objects.has_active_banner()
+        has_active_banners: bool = GlobalBanner.objects.has_active_banners()
```
A-Ashiq, line 21: Same comment as before. I'm not sure we need to use `has_active_banners()` anymore. That was only ever used as a validation check to enfore a single active banner at any given time, but that's no longer a restriction we should be enforcing

tests/unit/cms/snippets/models/test_global_banner.py
```diff
@@ -108,99 +107,3 @@ def test_inactive_global_banner_produces_correct_dunder_str(self):
         current_active_banner_label = "Active information-level global banner"
         assert current_active_banner_label not in string_representation
         assert "Inactive" in string_representation
-
-    @mock.patch.object(GlobalBannerManager, "has_active_banner")
```
A-Ashiq, line 112: These tests are fine to be removed 👍

tests/unit/cms/snippets/serializers/test_global_banner.py
```diff
@@ -119,5 +179,5 @@ def test_serialized_data_for_no_active_global_banner(self):
         )
 
         # Then
-        expected_data = {"active_global_banner": None}
+        expected_data = {"active_global_banners": []}
```
A-Ashiq, line 182: You need to update the test docstring to reference the empty list as the expected return value, no longer None

cms/snippets/serializers/global_banner.py
A-Ashiq, line 66: **Nitpick**: the plurality of this variable leads the reader to think they're getting back multiple objects. But in this case we're only getting either 1 or None

luketowell: The reason I pulled it out of the manager is so that I can change the manager to return the array and then we can temporarily manipulate the array in the serializer until we get rid of him.

cms/snippets/managers/global_banner.py
luketowell: No, I have now removed it and updated the tests.

tests/fakes/managers/cms/global_banner_manager.py
luketowell: 👍

cms/snippets/serializers/global_banner.py
luketowell: Good catch, this has been resolved

cms/snippets/serializers/global_banner.py
luketowell: Updated now due to adding the initial endpoint back in to the application.

tests/unit/cms/snippets/serializers/test_global_banner.py
luketowell: Updated 👍

tests/integration/cms/snippets/managers/test_global_banner.py
luketowell: Updated

cms/snippets/serializers/global_banner.py
luketowell: Updated

cms/snippets/serializers/global_banner.py
luketowell: Updated

cms/snippets/serializers/global_banner.py
luketowell: Updated

cms/snippets/views/global_banner.py
```diff
@@ -37,3 +38,30 @@ def get(cls, request, *args, **kwargs) -> Response:
         )
```
A-Ashiq, line 38: You will need to update the docstring in the v2 view banner
Screenshot 2025-03-11 at 16 14 26

tests/integration/cms/snippets/managers/test_global_banner.py
```diff
@@ -6,7 +6,7 @@
 
 class TestGlobalBannerManager:
     @pytest.mark.django_db
-    def test_has_active_banner(self):
+    def test_get_active_banners(self):
```
A-Ashiq, line 9: This test needs a re-write. The docstring still talks about returning True from when we checked banners existed.
But the assertion also is now checking its not None. But it will never be None. In the false condition it will be empty list

A-Ashiq:
@luketowell you will also need to update the help text which gets rendered on the CMS for the `is_active` field on the GlobalBanner snippet model:
cms/dynamic_content/help_texts.py

cms/dynamic_content/help_texts.py
```diff
@@ -297,9 +297,8 @@
 
 GLOBAL_BANNER_IS_ACTIVE: str = """
 Whether to activate this banner across every page in the dashboard. 
-Note that only 1 global banner can be active at a time.
-To switch from 1 active banner straight to another, 
-you must deactivate the 1st banner and save it before activating and saving the 2nd banner.
+Multiple banners can be active across the dashboard.
+Note: Warning banners are prioritised across the dashboard and will be displayed above information banners.
```
A-Ashiq, line 301: Out of curiosity how would we prioritise banners within each subgroup.
E.g. if we have 2 warning banners and another information banner all active, we would have:
```
Warning
Warning 
Information
```
But how do we decide the order of the warnings? Is this going to be handled by the frontend?

luketowell: This is a really good question, the ordering of warning and information groupings is straightforward however if we want to order those groupings by created date or something similar then we will need to add a field to the model so that we can subsequently order by that date time field.

PR #2328, opened 2025-03-31
Feature/cdd 2504 single page announcements

luketowell:
# Description

- The addition of an announcement panel to each of the page types. This enables single page announcements to be created and returned as part of the CMS/API.

Fixes #CDD-2504

---

## Type of change

---

# Checklist:


cms/common/models.py
```diff
@@ -82,3 +93,55 @@ class CommonPageRelatedLink(Orderable):
         APIField("url"),
         APIField("body"),
     ]
+
+
+class CommonPageAnnouncement(Orderable):
```
A-Ashiq, line 98: Perhaps we should move away from having supplementary models being on a per-page-type basis. And just have 1 `Announcement` type model that we can tie to any page type?

cms/common/migrations/0010_remove_commonpageannouncement_badge.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 5.1.7 on 2025-04-29 09:45
```
A-Ashiq, line 1: Can you condense all the migrations in each django app in your branch please? We should only need the 1 new migration in each of common, composite, home, metrics_documentation, topic, whats_new

cms/home/models/landing_page.py
```diff
@@ -1,11 +1,16 @@
...
+from wagtail.fields import RichTextField
+from wagtail.models import Orderable, Page
 
-from cms.dashboard.models import UKHSAPage
+from cms.dashboard.models import AVAILABLE_RICH_TEXT_FEATURES, UKHSAPage
+from cms.dynamic_content import help_texts
 from cms.dynamic_content.access import ALLOWABLE_BODY_CONTENT_SECTION_LINK
+from cms.dynamic_content.announcements import Announcement
 from cms.home.managers import LandingPageManager
+from cms.snippets.models.global_banner import BannerTypes
```
A-Ashiq, line 13: This is imported in a lot of places and doesn't look like it's being used anymore

cms/forms/help_texts.py
```diff
@@ -10,3 +10,21 @@
 CONFIRMATION_BODY_FIELD = """
 The main body of text to place on the post-submission page.
 """
+ANNOUNCEMENT_BANNER_TITLE: str = """
```
A-Ashiq, line 13: Why do we need to redefine these help texts in the form module? Is this a hangover from the previous solution?

luketowell: And like that... the dirt is gone

luketowell:
Updated the single page announcements so that they are now available from the base UKHSAPage. 
- filtering also works with the ability to only return active announcements 
- Announcements are prioritised to ensure Warning Announcements are displayed above Information Announcements: 

image

cms/dashboard/models.py
```diff
@@ -149,3 +153,11 @@ def get_url_parts(self, request=None) -> tuple[int, str, str]:
     @property
     def last_updated_at(self) -> datetime.datetime:
         return self.last_published_at
+
+    @property
+    def active_announcements(self):
```
A-Ashiq, line 158: Can you please add the return type hint for this

luketowell: Done

A-Ashiq (approved):

PR #2361, opened 2025-04-08
Task/cleanup redundant auth components/cdd 2531

A-Ashiq:
# Description

- Removes the now redundant `permissions` module as this will now be pushed down into the data layer in the model managers

Fixes #CDD-2531

---

## Type of change

---

# Checklist:


metrics/api/decorators/auth.py
```diff
@@ -16,9 +16,9 @@ def wrap(self, request, *args, **kwargs):
         try:
             group_id: str = request.headers[RBAC_AUTH_X_HEADER]
         except KeyError:
-            pass
-        else:
-            _set_rbac_group_permissions(request=request, group_id=group_id)
+            group_id = ""
```
A-Ashiq, line 19: By setting an empty string here and always calling `_set_rbac_group_permissions()` no matter what, we are making sure that the `request` object is always set with a list of permissions. 
So when we later access the `permissions` attribute on the request, we will always get a list back. In the case where the group id header was never included in the first place, we'll simply set an empty list of permissions.

This means we won't have to treat a None scenario differently. We can just consume that request later and loop through the permissions to determine if the user has access to the private component of that data. 

In this case, there will simply be an empty list and we'll just serve the user the public component of the data only

phill-stanley (approved):

PR #2366, opened 2025-04-09
Task/re implement fluent permissions/cdd 2531

A-Ashiq:
# Description

- Reimplements the `FluentPermissions` class so that it is able to wildcard match for any combination of parameters

Fixes #CDD-2531

---

## Type of change

---

# Checklist:


metrics/api/permissions/fluent_permissions.py
```diff
@@ -0,0 +1,80 @@
...
+            "age",
+            "stratum",
+        ]
+
+        for field in fields:
+            supporting_model = getattr(rbac_permission, field)
+            requested_value: str = getattr(self.requested_data_parameters, field)
+            if (
+                supporting_model is not None
+                and requested_value != supporting_model.name
```
A-Ashiq, line 57: Here we're saying, if the permission specifed the value for a supporting model i.e. a `Topic` or a `Theme` etc.
Then the requested value should match that supporting model name. 
So if I asked for `theme=infectious_disease` then the `Theme` model that I fetch via the `RBACPermission` should have the name `infectious_disease`. If False, then we don't have a match and we can assume.

Note that if the supporting model is `None` as in the `RBACPermission` didn't specify that, then this is a wildcard for that field. And as such, we don't return False for that field check.

metrics/api/permissions/fluent_permissions.py
```diff
@@ -0,0 +1,80 @@
...
+            ):
+                return False
+
+        return True
```
A-Ashiq, line 61: At this point we've exhausted all the fields and we have the following:
- Either a field was left as `None`
- The field was matched to the supporting model declared by the `RBACPermission`
Which in turn, means our `RBACPermission` model is allowing access to the requested dataset

tests/fakes/factories/metrics/rbac_models/rbac_permission.py
```diff
@@ -20,43 +24,75 @@ class Meta:
...
+        age: str = "",
+        stratum: str = "",
     ) -> FakeRBACPermission:
-        metric: FakeMetric = FakeMetricFactory.build_example_metric(
-            metric_name=metric_name,
-            topic_name=topic_name,
-            theme_name=theme_name,
-            sub_theme_name=sub_theme_name,
-        )
+        if theme:
```
A-Ashiq, line 36: This wall of gross looking if/else conditions allows us to more easily mix and match the fields on the permission

phill-stanley (approved):

PR #2572, opened 2025-07-18
Migrate away from `ListBlock` in favour of `StreamBlock` + `StructBlock`

A-Ashiq:
# Description

- Migrates us away from `ListBlock` in favour of the `StreamBlock` + `StructBlock` pattern

---

## Type of change

---

# Checklist:

### DIFF
--- a/cms/dynamic_content/global_filter/filter_types/data.py
+++ b/cms/dynamic_content/global_filter/filter_types/data.py
@@ -124,14 +124,20 @@ class DataCategorySelectionElement(blocks.StructBlock):
         icon = "form"
 
 
+class DataFilterElements(blocks.StreamBlock):
+    data_filter = DataFilterElement()
+
+
+class CategoryFilters(blocks.StreamBlock):
+    category = DataCategorySelectionElement()
+
+
 class DataFilters(blocks.StructBlock):
-    data_filters = blocks.ListBlock(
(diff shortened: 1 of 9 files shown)


cms/dynamic_content/global_filter/filter_types/data.py
```diff
@@ -24,7 +24,7 @@ def make_parameter_field_element(
 ) -> blocks.StructBlock:
     return blocks.StructBlock(
         [
-            ("label", blocks.CharBlock(required=True)),
+            ("label", blocks.CharBlock(required=False)),
```
A-Ashiq, line 27: If a label is not provided, then we'll ask the FE to fallback to using the actual value, thus reducing the burden on the content author

cms/dynamic_content/sections.py
```diff
@@ -35,44 +34,6 @@ class ContentCards(StreamBlock):
         help_text=help_texts.FILTER_LINKED_TIME_SERIES_CHART_TEMPLATE,
     )
 
-    def clean(self, value):
```
A-Ashiq, line 38: This validation will have to be at the page level. Otherwise, it will only validate successfully if the maps/charts are put in the same section as the global filter @phill-stanley 
For now I've removed it, but I'll set up a follow up ticket to put this validation back in at page level

PR #2593, opened 2025-07-29
Task/add cache flush handler for keys in reserved namespace/cdd 2729

A-Ashiq:
# Description

- Moves the tests in `test_management.py` for the cache management into a dedicated module
- Implements `copy()` methods on the cache clients
- Removes the `clear()` methods
- Adds the reserved staging namespace prefix
- When the crawler is being used to write to the cache, for reserved keys, they are written to the reserved staging are (ns3)
- In the next PR, the cache flush process will handle the migration of ns3 staging keys into ns2.
- When a request is being made to the API for a reserved key, then we lazy load in the ns2 namespace.

Fixes #CDD-2729

---

## Type of change

---

# Checklist:


caching/private_api/decorators.py
```diff
@@ -101,28 +101,45 @@ def _retrieve_response_from_cache_or_calculate(
...
-        request=request, is_reserved_namespace=is_reserved_namespace
+    if should_force_write:
+        # In this case if we are writing to the reserved namespace
+        # then we actually want to write to the reserved staging namespace
+        # So that we can copy all the staging keys into the reserved namespace
+        # when all the keys are ready
+        cache_entry_key: str = cache_management.build_cache_entry_key_for_request(
+            request=request,
+            is_reserved_staging_namespace=is_reserved_namespace,
+            is_reserved_namespace=False,
```
A-Ashiq, line 116: If we are force writing, and the `is_reserved_namespace` is true, then we actually want to write to the staging area, hence the weird placement of these booleans

caching/private_api/decorators.py
```diff
@@ -101,28 +101,45 @@ def _retrieve_response_from_cache_or_calculate(
...
+        )
+        return _calculate_response_and_save_in_cache(
+            view_function, timeout, cache_management, cache_entry_key, *args, **kwargs
+        )
+
+    # In this case, we have a lazy-load request.
+    # So we want to fetch the key associated with the non-reserved or reserved namespaces only.
+    cache_entry_key: str = cache_management.build_cache_entry_key_for_request(
```
A-Ashiq, line 124: If we've got here, then we are straight up lazy loading the data, so we need direct access to the main namespaces, either non-reserved or reserved but not reserved staging

caching/private_api/decorators.py
```diff
@@ -101,28 +101,45 @@ def _retrieve_response_from_cache_or_calculate(
...
-        view_function, timeout, cache_management, cache_entry_key, *args, **kwargs
-    )
+    try:
+        return cache_management.retrieve_item_from_cache(
+            cache_entry_key=cache_entry_key
+        )
+    except CacheMissError:
+        # If it is not in the cache then we'll write it
+        # on the way out after calculating the response
+        return _calculate_response_and_save_in_cache(
```
A-Ashiq, line 137: Even if it's a reserved key, if we've got here then we've ended up in a situation whereby the request has had to punch through the cache, because the key wasn't in the reserved namespace. So here we'll just lazy load it into the cache on our way back from the server

caching/private_api/management.py
```diff
@@ -123,15 +145,6 @@ def save_item_in_cache(
         self._client.put(cache_entry_key=cache_entry_key, value=item, timeout=timeout)
         return item
 
-    def clear(self) -> None:
```
A-Ashiq, line 126: We shoudn't ever now need to empty the cache completely so we can get rid of this

phill-stanley (approved):

PR #2694, opened 2025-09-04
Task/implement dual redis cache strategy/cdd 2839

A-Ashiq:
# Description

This is the 1st PR to implement the dual cache strategy.
With this new approach we will be:
- Seperating ephemeral data (charts, tables, downloads etc) from long lived / reserved data (maps) into physically seperate caches - they are currently in 2 namespaces within the same cache.
- Simplifying how we flush caches to be naive:
    - Empty the selected cache completely with the `clear()` which wraps around the `FLUSHDB` command
    - Crawl the app and lazy load the responses into the cache
    - Reconfigure the `CacheManagement` and underlying `CacheClient` so that when they are getting or putting items in the cache, they select the cache based on the key. If it begins with `ns2-` it goes in the reserved cache otherwise the default cache
    - Simplify the `cache_response()` decorator so that it is a simple lazy-load of the request into the cache. Its the responsibility of the previous clear cache command to ensure the cache has been cleared so the next version of data can be persisted.

- Adds the connection for the django app to the `reserved` cache
- Removes reference to `force_refresh`, previously when we were crawling the app, we'd set the cache crawler to always punch through the cache and write. But we don't really need this if we empty the selected cache ahead of time.

Fixes #CDD-2839

---

## Type of change

---

# Checklist:

### DIFF
--- a/caching/internal_api_client.py
+++ b/caching/internal_api_client.py
@@ -41,7 +41,6 @@ class InternalAPIClient:
         self,
         *,
         client: APIClient | None = None,
-        force_refresh: bool = False,
         reserved_namespace: bool = False,
     ):
         self._client = client or self.create_api_client()
@@ -60,7 +59,6 @@ class InternalAPIClient:
         self.menus_endpoint_path = MENUS_ENDPOINT_PATH
 
         # Header configurations
-        self.force_refresh = force_refresh
         self.reserved_namespace = reserved_namespace
 
     # API client
@@ -79,7 +77,6 @@ class InternalAPIClient:
 
     def build_headers(self) -> dict[str, bool]:
         return {
-            CACHE_FORCE_REFRESH_HEADER_KEY: self.force_refresh,
             CACHE_RESERVED_NAMESPACE_HEADER_KEY: self.reserved_namespace,
         }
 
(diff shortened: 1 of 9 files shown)


caching/internal_api_client.py
```diff
@@ -41,7 +41,6 @@ def __init__(
         self,
         *,
         client: APIClient | None = None,
-        force_refresh: bool = False,
```
A-Ashiq, line 44: We'll no longer let cache crawlers force write to the cache. We'll clear the cache first and then crawl the app. If another request makes it to the server first then that will just lazy load into the cache and the crawler will then just bounce back from the cache with the pre-computed response and move onto the next item

tests/unit/caching/private_api/test_decorators.py
```diff
@@ -167,243 +161,6 @@ def test_item_not_cached_when_caching_v2_enabled_is_set_to_true(
         assert retrieved_response == spy_calculate_response_from_view.return_value
 
 
-class TestForceCacheRefreshHeader:
```
A-Ashiq, line 170: All these tests are about how we handle force writes to the cache, but we'll no longer support this in order to keep things simple

phill-stanley (approved):

PR #2695, opened 2025-09-04
CDD-2839: Implement dual redis cache strategy - Part 2

A-Ashiq:
# Description

This is the 2nd PR to implement the dual cache strategy.

- Renames the constructor method `create_crawler_to_force_write_in_non_reserved_namespace()` on the `PrivateAPICrawler` -> `create_crawler_for_default_cache()`
- Renames the constructor method `create_crawler_to_force_write_in_reserved_namespace()` on the `PrivateAPICrawler` -> `create_crawler_for_reserved_cache()`
- The above renaming was done because we will no longer force write requests. They'll be lazy loaded into the cache relying on the fact that the selected cache has been cleared ahead of time.

Fixes #CDD-2839

---

## Type of change

---

# Checklist:

### DIFF
--- a/caching/private_api/crawler/private_api_crawler.py
+++ b/caching/private_api/crawler/private_api_crawler.py
@@ -76,7 +76,7 @@ class PrivateAPICrawler:
         return cls(internal_api_client=internal_api_client)
 
     @classmethod
-    def create_crawler_to_force_write_in_reserved_staging_namespace(cls) -> Self:
+    def create_crawler_for_reserved_cache(cls) -> Self:
         internal_api_client = InternalAPIClient(reserved_namespace=True)
         return cls(internal_api_client=internal_api_client)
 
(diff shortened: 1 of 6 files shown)


caching/private_api/crawler/private_api_crawler.py
```diff
@@ -71,12 +71,12 @@ def __init__(
     # Class constructors
 
     @classmethod
-    def create_crawler_to_force_write_in_non_reserved_namespace(cls) -> Self:
+    def create_crawler_for_default_cache(cls) -> Self:
```
A-Ashiq, line 74: The crawlers will end up being quite simple in structure as they won't be able to force write.
We empty the cache completely ahead of time and crawl the app with these crawlers. If they come across an item which has been persisted in the time between the cache being emptied and the cache crawler getting to that item, then that's fine there's no need to force write as we know that item has been freshly computed from the origin

phill-stanley (approved):

PR #2696, opened 2025-09-04
CDD-2839: Implement dual redis cache strategy - Part 3

A-Ashiq:
# Description

This is the 3rd PR to implement the dual cache strategy.

- When processing bulk downloads we will now re-use the same cache crawler as the cache crawler which we use to warm the default cache as this is simple lazy loading of requests

Fixes #CDD-2839

---

## Type of change

---

# Checklist:

### DIFF
--- a/caching/private_api/handlers.py
+++ b/caching/private_api/handlers.py
@@ -166,5 +166,5 @@ def get_all_downloads(*, file_format: str = "csv") -> list[dict[str, str]]:
 
     """
     pages = collect_all_pages()
-    crawler = PrivateAPICrawler.create_crawler_for_lazy_loading()
+    crawler = PrivateAPICrawler.create_crawler_for_default_cache()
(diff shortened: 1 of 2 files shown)


caching/private_api/handlers.py
```diff
@@ -166,5 +166,5 @@ def get_all_downloads(*, file_format: str = "csv") -> list[dict[str, str]]:
 
     """
     pages = collect_all_pages()
-    crawler = PrivateAPICrawler.create_crawler_for_lazy_loading()
+    crawler = PrivateAPICrawler.create_crawler_for_default_cache()
```
A-Ashiq, line 169: When we are gathering all the downloads on every page for the bulk downloads functionality. We can just re-use the normal lazy-load crawler. Technically it wouldn't matter which crawler we use here because the outbound request would just get picked up the `CacheClient` and routed to the correct cache anyway

phill-stanley (approved):

PR #2700, opened 2025-09-04
CDD-2839: Implement dual redis cache strategy - Part 7

A-Ashiq:
# Description

This is the 7th PR to implement the dual cache strategy.

- Simplifies the `cache_response()` decorator which we use to cache all our private API endpoints by:
    - Removing support for force cache writes
    - Removing support for reserved staging keys
    - Re-implementing simple lazy loading. If the key is in the cache take it from there otherwise calculate and write it on the way out
    - Let the `CacheManagement` and underlying `CacheClient` decide which cache the current item should be written to - which is based on the cache key
    - Note that currently this is dictated by how we use the `cache_response()` decorator. E.g. if we say `cache_response(is_reserved_namespace=True)` as is the case for the maps API, then all maps requests will be put in the reserved cache

Fixes #CDD-2839

---

## Type of change

---

# Checklist:

### DIFF
--- a/caching/private_api/decorators.py
+++ b/caching/private_api/decorators.py
@@ -4,10 +4,6 @@ from functools import wraps
 from rest_framework.request import Request
 from rest_framework.response import Response
 
-from caching.internal_api_client import (
-    CACHE_FORCE_REFRESH_HEADER_KEY,
-    CACHE_RESERVED_NAMESPACE_HEADER_KEY,
-)
 from caching.private_api.management import CacheManagement, CacheMissError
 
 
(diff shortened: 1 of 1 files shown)


caching/private_api/decorators.py
```diff
@@ -102,27 +87,13 @@ def _retrieve_response_from_cache_or_calculate(
     if is_caching_v2_enabled():
```
A-Ashiq, line 87: When we switch to caching v2 i.e. leaning on Next.js's in-built caching, we might want to pad out this conditional. Something like:
```
if caching_v2_enabled and not `is_reserved_namespace`
``
That way we can continue to keep the long-lived data in a seperate cache, whilst leaving all the ephemeral data to be taken care of by Next.js

PR #2701, opened 2025-09-04
CDD-2839: Implement dual redis cache strategy - Part 8

A-Ashiq:
# Description

This is the 8th PR to implement the dual cache strategy.

- Makes the `CacheClient` class take a `is_reserved_namespace` boolean to decide which cache to preselect, used in the case of `clear()` commands
- Implements a `_select_cache_for_key()` on the `CacheClient` which picks the target cache based on the given cache key, used in the case of `get()` and `put()` commands when dealing with individual items in the cache
- Removes the `list_keys()`, `delete_many()` & `copy()` commands from the `CacheClient` class which were only being used when doing the blue-green approach for the reserved cache, but since we're now gonna simplify that we'll no longer use those commands - they also won't work as we want them to in production without making the application code cluster aware which I thought was needless complexity.
- Implements a `clear()` method on the `CacheClient`, this calls `clear()` on the pre-selected cache which is just a wrapper around the `FLUSHDB` command. In AWS Elasticache Redis Serverless, each node only considers the DB at #0 and we know that the `FLUSHDB` command will be respected by AWS and will be propagated to all shards in the cluster regardless of how far out the cluster has scaled
- Removes `clear_non_reserved_keys()`, `delete_many()`, `get_reserved_staging_keys()` & `move_all_reserved_staging_keys_into_reserved_namespace()` & `get_reserved_keys()` from `CacheManagement` as these will also no longer be used
- Reimplements the `force_cache_refresh_for_reserved_namespace()` handler to clear the reserved cache and then simply warm the reserved cache without dealing with the staging area approach

Fixes #CDD-2839

---

## Type of change

---

# Checklist:

### DIFF
--- a/tests/unit/caching/private_api/test_client.py
+++ b/tests/unit/caching/private_api/test_client.py
@@ -232,3 +232,20 @@ class TestInMemoryCacheClient:
 
         # Then
         assert retrieved_value == mocked_value
+
+    def test_clear_empties_the_cache(self):
+        """
+        Given a cache with an existing key
+        When `clear()` is called from an instance of the `InMemoryCacheClient`
+        Then the cache is emptied
+        """
+        # Given
+        fake_cache = {"abc": mock.Mock()}
+        in_memory_cache_client = InMemoryCacheClient()
+        in_memory_cache_client._cache = fake_cache
+
+        # When
+        in_memory_cache_client.clear()
+
+        # Then
+        assert not in_memory_cache_client._cache


caching/private_api/client.py
```diff
@@ -51,75 +73,37 @@ def put(self, *, cache_entry_key: str, value: Any, timeout: int | None) -> None:
             None
 
         """
-        self._cache.set(key=cache_entry_key, value=value, timeout=timeout)
-
-    def list_keys(self) -> list[bytes]:
```
A-Ashiq, line 56: We no longer need to list keys, copy them, or move them in any way. So we can safely get rid of all the methods which comprise those operations

caching/private_api/handlers.py
```diff
@@ -135,24 +137,25 @@ def force_cache_refresh_for_reserved_namespace(
...
         geographies_api_crawler=private_api_crawler.geography_api_crawler
     )
 
-    original_reserved_keys: list[str] = cache_management.get_reserved_keys()
     crawl_all_pages(
         private_api_crawler=private_api_crawler,
         area_selector_orchestrator=area_selector_orchestrator,
     )
-    cache_management.delete_many(keys=original_reserved_keys)
-    cache_management.move_all_reserved_staging_keys_into_reserved_namespace()
```
A-Ashiq, line 155: Now that we'll just empty the cache and then crawl the app, we don't need to do list, copy, move type operations - which were all tied to the AWS limitation in shard-specific commands in clusters.

phill-stanley (approved):

PR #2715, opened 2025-09-10
Task/insert null values for subplots when missing/cdd 2788

A-Ashiq:
# Description

- Adds `null` case placeholders for any plots with missing data so that the frontend can consume the API response and consistently add `-` placeholders for those rows with missing data

Fixes #CDD-2788

---

## Type of change

---

# Checklist:

### DIFF
--- a/metrics/interfaces/tables/subplot_tables/access.py
+++ b/metrics/interfaces/tables/subplot_tables/access.py
@@ -23,8 +23,10 @@ class SubplotTablesInterface:
         result: list[dict] = []
 
         for request_params in request_params_per_group:
-            tabular_data_for_sub_plot = self._generate_complete_tabular_data_for_subplot(
-                request_params=request_params
+            tabular_data_for_sub_plot = (
+                self._generate_complete_tabular_data_for_subplot(
+                    request_params=request_params
+                )
             )
             result += tabular_data_for_sub_plot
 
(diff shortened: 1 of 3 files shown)


metrics/interfaces/tables/subplot_tables/access.py
```diff
@@ -19,23 +20,106 @@ def generate_full_plots_for_table(self) -> list[dict]:
...
 
         for request_params in request_params_per_group:
-            try:
-                tabular_data_for_sub_plot: list[dict[str, str]] = (
-                    generate_table_for_full_plots(request_params=request_params)
+            tabular_data_for_sub_plot = (
+                self._generate_complete_tabular_data_for_subplot(
+                    request_params=request_params
                 )
-            except (InvalidPlotParametersError, DataNotFoundForAnyPlotError):
```
A-Ashiq, line 30: Because of the new `_inject_null_values_for_any_missing_plots()` method, we'll now always return a 200. It will just contain null value cases everywhere instead.
The logic of this all feels simpler by just calling out to `_inject_null_values_for_any_missing_plots()` instead of then having to do another check after the fact to say hey if everything was null then throw the error and return a 400.

For a public/user facing API I'd say this is a bad pattern, but given this sits on the private and only the frontend consumes this endpoint I think it's fine

PR #3077, opened 2026-03-17
task/CDD 1379 page previews

ghost:
# Description

## CDD-1379 - Page Previews

**Depends Upon**:  Front End branch: [task/CDD-1379-page-previews](https://github.com/UKHSA-Internal/data-dashboard-frontend/pull/876) (please pull the latest)

**Date:** 2026-02-27

**Ticket:** https://ukhsa.atlassian.net/browse/CDD-1379?search_id=055fe61d-bee9-48d9-80bc-ffb0f1c26b76&referrer=quick-find

**Authors:** Jean-Pierre Fouche

**Impact:** Affects all pages - broad testing required

**Testing:** Comprehensive unit tests supplied.  UAT needed.

## Summary

Allow editors of headless composite pages to click a **Preview** button that immediately redirects them to the external frontend application, rather than opening the built-in Wagtail iframe preview.  Preview URLs include a short-lived signed token so the frontend can safely fetch draft content from the CMS. 

Additionally, allow users to select an embargo date in order to preview otherwise embargoed data.

Continued in ./changelog/2026-02-27/CDD-1379.page-previews.md

Fixes #CDD-1379

---

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

---

# Checklist:

- [X] I have made corresponding changes to the documentation
- [X] I have added screenshots or screen grabs where appropriate


cms/dashboard/viewsets.py
```diff
@@ -41,42 +50,75 @@ def get_queryset(self):
...
     def detail_view(self, request: Request, pk: int) -> Response:
         """This end point returns a page from the CMS based on a Page `ID`."""
         return super().detail_view(request=request, pk=pk)
 
 
 @extend_schema(tags=["cms"])
-class CMSDraftPagesViewSet(PagesAPIViewSet):
-    base_serializer_class = CMSDraftPagesSerializer
-    permission_classes = []
+class CMSPagesAPIViewSet(BaseCMSPagesAPIViewSet):
```
jeanpierrefouche-ukhsa: API route for published Pages.  Note caching is enabled by means of the decorator.

cms/dashboard/views.py
```diff
@@ -1,6 +1,135 @@
...
+        page = get_object_or_404(Page, pk=pk).specific
+
+        perms = page.permissions_for_user(request.user)
+        if not perms.can_edit():
+            raise PermissionDenied
+
+        now = timezone.now()
+        payload = {
+            "page_id": page.pk,
+            "editor_id": request.user.pk if request.user.is_authenticated else None,
```
jrdh: Two questions:
- What is this `editor_id` key used for?
- The `if request.user.is_authenticated` check is presumably redundant because of the `if not perms.can_edit()` line above surely?

cms/dashboard/views.py
```diff
@@ -1,6 +1,135 @@
...
+        # See docs/environment_variables.md for PAGE_PREVIEWS_FRONTEND_URL_TEMPLATE format.
+        template = getattr(
+            settings,
+            "PAGE_PREVIEWS_FRONTEND_URL_TEMPLATE",
+            "http://localhost:3000/preview?slug={slug}&t={token}",
+        )
+
+        route_slug = self.build_route_slug(page=page)
+
+        frontend_url = template.format(
```
jrdh: I've not seen anywhere where slug_name and page_id are actually in the template? Are they included just to be as flexible as possible?

cms/dashboard/viewsets.py
```diff
@@ -41,42 +50,75 @@ def get_queryset(self):
...
+            return Response(status=status.HTTP_404_NOT_FOUND)
+
+        # Ensure token page_id matches
+        payload_page_id = payload.get("page_id")
+        if payload_page_id is None or int(payload_page_id) != int(instance.pk):
+            return Response(status=status.HTTP_401_UNAUTHORIZED)
+
+        # Return latest draft if available, else published
+        latest_revision = instance.get_latest_revision()
+        if latest_revision:
```
jrdh: This could be a bit DRYer in my books:

```python
latest_revision = instance.get_latest_revision()
if latest_revision:
    draft_page = latest_revision.as_object()
    serializer = self.get_serializer(draft_page)
else:
    serializer = self.get_serializer(instance)
return Response(serializer.data)
```
or you could even:
```python
latest_revision = instance.get_latest_revision()
if latest_revision:
    instance = latest_revision.as_object()
serializer = self.get_serializer(instance)
return Response(serializer.data)
```
though I'm less of a fan of that because it splats the instance variable with a different meaning.

cms/dashboard/wagtail_hooks.py
```diff
@@ -1,14 +1,104 @@
...
+from wagtail.admin.widgets import Button
 from wagtail.models import Page
 from wagtail.whitelist import check_url
 
+from cms.dashboard.views import PreviewToFrontendRedirectView
+
+VIEW_LIVE_LABEL = "View Live"
+
+
+# Move FrontendPreviewAction to module level
```
jrdh: Not sure I understand why this comment is here? Was it a todo?

cms/dashboard/wagtail_hooks.py
```diff
@@ -1,14 +1,104 @@
...
+
+
+def _patched_create_view_live_message_button(self):
+    live_url = _build_view_live_url(page=self.page)
+    return messages.button(live_url, _("View live"), new_window=True)
+
+
+def _patched_edit_view_live_message_button(self):
+    live_url = _build_view_live_url(page=self.page)
+    return messages.button(live_url, _("View live"), new_window=True)
```
jrdh: Given this and the function above do excatly the same thing I think they should just be one function. If they need to be separated to do different jobs later then they can be but them being separated now when they do the same thing is more likely to result in maintenance errors down the line.

cms/dashboard/wagtail_hooks.py
```diff
@@ -119,8 +278,93 @@ def _get_page_url(page_id: int) -> str:
...
+    Helper to insert the FrontendPreviewAction for the 'Preview' button.
+    Extracted to ensure explicit branch coverage.
+    """
+    try:
+        admin_url = reverse("cms_preview_to_frontend", args=[page.pk])
+    except NoReverseMatch:
+        template = getattr(
+            settings,
+            "PAGE_PREVIEWS_FRONTEND_URL_TEMPLATE",
+            "http://localhost:3000/preview?slug={slug}",
```
jrdh: This template default is different to the one in `PreviewToFrontendRedirectView.get` as is some of the logic that comes after it when I think it should be common and shared. `build_route_slug` already is but the surrounding parts look like they should really be the same but they are different.

docs/environment_variables.md
```diff
@@ -130,6 +130,42 @@ See the [django docs](https://docs.djangoproject.com/en/4.2/ref/settings/#std-se
 
 ---
 
+### CMS Page previews configuration
+
+Wagtail Content Management System (CMS) allows users to preview pages in the configured front-end.  The following environment variables are supported:
```
jrdh: I think it'd be worth adding into this that we have customised this behaviour because we're using a headless Wagtail instance to avoid confusion.

jrdh (changes requested):
I've made lots of comments but this is looking really good! Worked nicely on my local machine 🙂

cms/dashboard/wagtail_hooks.py
```diff
@@ -88,6 +178,75 @@ def register_icons(icons: list[str]) -> list[str]:
...
+        )
+        admin_url = template.format(
+            slug=PreviewToFrontendRedirectView.build_route_slug(page=page)
+        )
+
+    return [
+        Button(
+            label=button_label,
+            url=admin_url,
+            priority=10,
```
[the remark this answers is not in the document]
jeanpierrefouche-ukhsa: Icon added!

metrics/data/managers/core_models/time_series.py
```diff
@@ -509,6 +510,27 @@ def find_latest_released_embargo_for_metrics(
 class CoreTimeSeriesManager(models.Manager):
     """Custom model manager class for the `TimeSeries` model."""
 
+    @staticmethod
+    def _has_access_to_non_public_data(
```
jeanpierrefouche-ukhsa: Extracts the inline validate_permissions_for_non_public() call from query_for_data into a dedicated _has_access_to_non_public_data static method.

The logic is identical to main. It is a Sonar "Extract Method" refactoring to reduce the cognitive complexity of query_for_data. The violation pre-existed on main and has been fixed here as a by-product of touching this file.

scripts/_quality.sh
```diff
@@ -57,7 +57,7 @@ function _quality_format_check() {
     if [ -n "${changed_files}"  ]; then
       echo "Some files appear to be un-formatted. Please rectify by running 'uhd quality format' and committing the changes."
       echo "${changed_files}"
-      exit 1
+            return 1
```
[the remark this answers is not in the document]
jeanpierrefouche-ukhsa: Yes. Job done!

cms/dashboard/management/commands/build_cms_site.py
```diff
@@ -49,8 +50,10 @@ def handle(self, *args, **options):
         wagtail_root_page.add_child(instance=root_page)
         wagtail_root_page.save_revision().publish()
 
+        frontend_url = os.environ.get("FRONTEND_URL", "localhost")
+
         Site.objects.create(
-            hostname=os.environ.get("FRONTEND_URL", "localhost"),
+            hostname = urlparse(frontend_url).hostname or "localhost",
```
ghost: Bug fix: need to extract the hostname, and not assume that FRONTEND_URL is a hostname.

common/virtual_clock.py
```diff
@@ -0,0 +1,100 @@
...
+This module provides a virtual clock that allows authorised preview users to
+view data as if the current time were set to the future (or past).
+It is used to support embargo previews, by enabling the consumer to view data
+that is presently restricted.
+"""
+
+import contextvars
+import datetime
+import logging
+import typing as t
```
jrdh: Given this is only used once in this file I think this should just be `from typing import Any`.

common/virtual_clock.py
```diff
@@ -0,0 +1,100 @@
...
+    else:
+        return None
+
+    try:
+        return datetime.datetime.fromtimestamp(epoch_seconds, tz=datetime.UTC)
+    except (OverflowError, OSError, ValueError):
+        return None
+
+
+def set_embargo_time(embargo_time_value: object, *, token: str) -> bool:
```
jrdh: Any particular reason to force the token to be a kw but not the embargo_time_value?

common/virtual_clock.py
```diff
@@ -0,0 +1,100 @@
...
+    Falls back to timezone.now().
+    """
+    embargo_time = _embargo_time_ctx.get()
+    if isinstance(embargo_time, datetime.datetime):
+        return embargo_time
+
+    return timezone.now()
+
+
+def get_embargo_time_context() -> contextvars.ContextVar[datetime.datetime | None]:
```
jrdh: This is only used in tests, is it really necessary?

common/virtual_clock.py
```diff
@@ -0,0 +1,100 @@
...
+            return None
+    elif isinstance(embargo_time_value, (int, float)) and not isinstance(
+        embargo_time_value, bool
+    ):
+        epoch_seconds = int(embargo_time_value)
+    else:
+        return None
+
+    try:
+        return datetime.datetime.fromtimestamp(epoch_seconds, tz=datetime.UTC)
```
jrdh: I assume UTC is ok here? Should we use the tz django is using instead to be consistent? Which one gets used for embargo filtering in the actual database query?

common/virtual_clock.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def set_embargo_time(embargo_time_value: object, *, token: str) -> bool:
+    """Set embargo time for current request context after validation.
+
+    The value must be either `now` or valid epoch seconds.
+    """
+    page_previews_enabled = getattr(settings, "PAGE_PREVIEWS_ENABLED", False)
+    if not page_previews_enabled:
+        _embargo_time_ctx.set(timezone.now())
```
jrdh: Why does this default the time to now if page previews is off instead of just setting to None so that now is used at request time?

metrics/api/settings/default.py
```diff
@@ -296,16 +301,52 @@
...
+        urlsafe_b64encode(hmac.new(secret_key, namespace_primary, sha512).digest())
+        .decode("ascii")
+        .rstrip("=")
+    )
+    secondary = (
+        urlsafe_b64encode(hmac.new(secret_key, namespace_secondary, sha512).digest())
+        .decode("ascii")
+        .rstrip("=")
+    )
+    return (primary + secondary)[:120]
```
jrdh, line 378: If we're doing this, why do we have a salt env var (`PAGE_PREVIEWS_TOKEN_SALT`) in the infra PR?

validation/shared.py
```diff
@@ -1,3 +1,94 @@
+import logging
```
jrdh: Not sure this file should live here but don't have a good suggestion of where it should go 😬. The validation module is used for _data_ validation whereas this is doing authentication type validation work. Also the name `shared.py` should be changed to something more useful.

jrdh (changes requested):
Still looking good. Made a few comments in line but nothing major there. I have concern which I think we have to fix and a question.

The concern is that I think the `/nocache/` functionality should be protected by the same method we're using for the `/preview/` URLs (i.e. signing) as without any authentication on them they allow access to potentially unembargoed data and open up endpoints malicious users could DDOS us with due to the uncached nature of the responses.

The question is whether you considered using the Django request context instead of the `ContextVar`s? I don't have a strong feeling either way but wasn't sure if there was a best practice method Django recommended for variables that live for the length of a request and need to be passed around to different functions? The only issues I could foresee is if Django actually uses multiple threads to respond to a request and doesn't copy the context over which would mean the embargo date wouldn't always be set correctly. I think this is really unlikely but just wanted to raise it as a concern related to what Django best practice is.

As an aside, I would have much preferred if this had been worked on in a series of commits rather than one fat one. It's not good for future maintainability/the git history, and it makes it much harder to review as I can't just look at the new commits created after my last review.

ghost:
> The question is whether you considered using the Django request context instead of the `ContextVar`s? I don't have a strong feeling either way but wasn't sure if there was a best practice method Django recommended for variables that live for the length of a request and need to be passed around to different functions? The only issues I could foresee is if Django actually uses multiple threads to respond to a request and doesn't copy the context over which would mean the embargo date wouldn't always be set correctly. I think this is really unlikely but just wanted to raise it as a concern related to what Django best practice is.

Hi Josh, 

Thanks for your comment.

Yes, Django request context was considered and I appreciate that it is often the go-to approach in Django.  However, Django request context needs to be passed around.  This would break all of the code interfaces (signatures) and make the application more complicated.  ContextVars allow the code to pull in request-scoped state, without passing state around.  A few points about ContextVars: 

- they are safe (won't leak) within the execution context (which is deeper than just the thread.  See Python Docs on ContextVar, the section on [asyncio support](https://docs.python.org/3/library/contextvars.html)
- if spawning a new thread, the onus is now placed upon the thread creator to use ContextVar.copy_context.  Note that this is an edge case, in which the thread creator needs access to embargo_time.   In my view, this is cleaner, relatively simpler,  and more reliable than passing Django request state around explicitly.  Anybody spawning a new thread should, in any case, reasonably assume that there are going to be issues with accessing state and take appropriate steps to ensure access.

Example code, below:

```python
import threading
import contextvars

embargo_date = contextvars.ContextVar("embargo_date")

def worker():
    # ContextVar is available here (current execution context, current thread)
    print(embargo_date.get())


def start_thread():
    embargo_date.set("2026-05-14")

    # Capture current context
    ctx = contextvars.copy_context() # <================== use copy_context if you need the embargo_date

    # Run worker inside copied context
    thread = threading.Thread(target=lambda: ctx.run(worker))  # <===== run a thread within the context
    thread.start()
    thread.join()
    ```

**Comment by ghost** [comment]
> The concern is that I think the `/nocache/` functionality should be protected by the same method we're using for the `/preview/` URLs (i.e. signing) as without any authentication on them they allow access to potentially unembargoed data and open up endpoints malicious users could DDOS us with due to the uncached nature of the responses.

Hi Josh, 

I have reworked the flow so that routes to "View Live" are now authenticated.  
Please see the changes for your review.

Many Thanks!

@jrdh

**Comment by ghost** [comment]
> As an aside, I would have much preferred if this had been worked on in a series of commits rather than one fat one. It's not good for future maintainability/the git history, and it makes it much harder to review as I can't just look at the new commits created after my last review.

Hi Josh, 

Yes, I can see where you are coming from! In a large commit like this it must have been less than helpful having to do the second review, not having a baseline to review it against.

@jrdh

File: common/virtual_clock.py (hunk 13 above)
Comment (ghost) [reply common/virtual_clock.py]: Yes.  Updated accordingly.

File: common/virtual_clock.py (hunk 14 above)
Comment (ghost) [reply common/virtual_clock.py]: Fixed this! :)

File: common/virtual_clock.py (hunk 15 above)
Comment (ghost) [reply common/virtual_clock.py]: Very good! :)
I have removed it!

File: common/virtual_clock.py (hunk 17 above)
Comment (ghost) [reply common/virtual_clock.py]: Indeed! 
I have set it to None.

File: metrics/api/settings/default.py (hunk 18 above)
Comment (ghost) [reply metrics/api/settings/default.py:378]: Agreed!  
I will remove the env var from the infra repo.

File: common/virtual_clock.py (hunk 16 above)
Comment (ghost) [reply common/virtual_clock.py]: UTC is OK here and is indeed recommended, to cover Daylight Saving time variance (GMT vs BST). 

The `django.utils` timezone would make it more consistent, as `django.utils` is being used in the ORM query.  I am therefore pushing the change to use django.utils instead of the built-in.  (Note, however, that the two are functionally equivalent and the change is not essential)

File: validation/shared.py (hunk 19 above)
Comment (ghost) [reply validation/shared.py]: Rearranged into the following modules: 

./common/request_caching.py
./common/page_previews.py

I left the one remaining function (which is not related to this PR) in validation/shared.py

File: cms/dashboard/views.py [hunk 20]
```diff
@@ -253,6 +263,14 @@ def get(self, request, pk):
             raise PermissionDenied
 
         embargo_time_value = request.GET.get("embargo_time")
+        route = request.GET.get("route", "nocache")
```
jrdh, line 266: Any particular functional reason to default to `nocache` here instead of error if the `route` parameter isn't set?

cms/dashboard/wagtail_hooks.py
```diff
@@ -256,16 +238,16 @@ def register_admin_urls():
     """Register admin URLs for CMS dashboard views.
 
     We register an admin redirect endpoint
-    (`/admin/preview-to-frontend/<pk>/`) that signs a short-lived
+    (`/admin/redirect-to-frontend/<pk>/`) that signs a short-lived
```
jrdh, line 241: Extremely minor note, because we expose `/admin` as `/cms-admin`, the URL path starts `cms-admin` not `admin`.

jrdh (approved):
Looks good, thanks for making those changes JP 🎉 

Couple of minor comments in line and a question:
- If I remove the `page_id` or change it to a different valid value (e.g. the ID of another page) from either `/preview` or `/nocache` URLs it still works fine - is this intended? If so do we need that query parameter?

Otherwise, this is good to go to QA as far as I'm concerned!

jrdh (approved):

jrdh (approved):

aidan (approved):

PR #3079, opened 2026-03-18
Task/cdd 3206 be   filter db query for only is public   true data

dandammann:
# Description

Before I started this ticket:

The data model itself is not restricted, so Kathryn's previous work on CDD-2983: BE: Set `is_public` flag to be mandatory field
Test Completed
  means we plan to ingest and store both is_public=True/False headline/time-series data in the near future.

The is_public flag only exists on headline/time-series row data. Access decisions are evaluated using request dimensions (theme, sub_theme, topic, metric, geography, geography_type) for those row queries, but this does not mean all metadata/master-table entities are globally filtered by is_public. Please evaluate whether this is acceptable, or if not, what should be done about that!

The codebase already contains RBAC (Role-Based Access Control) logic for headline/time-series data: non-authorized paths return public rows; authorized RBAC paths can include non-public rows when permitted.

New functionality I introduced:

I have now introduced an ENFORCE_PUBLIC_DATA_ONLY=True flag in /metrics/api/settings/auth.py that suppresses this RBAC functionality and only returns is_public=True dashboard API data. The following exceptions are intentional:

 a) The API endpoints /api/audit/* can still return public and non-public diagnostic data.
 b) Any API endpoints that serve metadata/master-table style data (not headline/time-series row payloads) remain unfiltered too.

If we later switch ENFORCE_PUBLIC_DATA_ONLY=False, the RBAC-based behaviour from point 3. above is re-enabled. The automation tests allow for the flag being either True/False, so will not have to be amended when the flag gets switched. Long-term all temporary enforcement code to do with ENFORCE_PUBLIC_DATA_ONLY should be removed, once the final application state and data flow has been achieved.

## Type of change

---

## Checklist:


metrics/api/settings/auth.py
```diff
@@ -5,4 +5,32 @@ def is_auth_enabled() -> bool:
...
 
+def is_allow_missing_is_public_field() -> bool:
+    return str(os.environ.get("ALLOW_MISSING_IS_PUBLIC_FIELD", "")).lower() in {
+        "true",
+        "1",
+    }
+
+
+def is_enforce_public_data_only() -> bool:
+    return str(os.environ.get("ENFORCE_PUBLIC_DATA_ONLY", "true")).lower() in {
```
kathryn-dale: How is this different to "AUTH_ENABLED"? Feels like a duplicate, as if auth is not enabled, only public data should be returned

public_api/serializers/api_time_series_request_serializer.py
```diff
@@ -78,7 +78,9 @@ def get_queryset(self) -> "APITimeSeriesQuerySet":
         """
         kwargs: dict[str, str] = self.get_kwargs_from_request()
         return self.api_time_series_manager.get_distinct_column_values_with_filters(
-            lookup_field=self.lookup_field, restrict_to_public=True, **kwargs
+            lookup_field=self.lookup_field,
+            restrict_to_public=True,  # because we are not allowing non-public data through the public API
```
kathryn-dale: Have we confirmed 100% that we are not allowing non-public data through the public api?

dandammann: Source-code analysis did not give me enough confidence to say that. Only testing did.

It may be good, if you also review my tests (see last ticket comment, and browse my "Private-API-SQL-Statements.txt" and "Public-API-SQL-Statements.txt" test results to see, if you have any concerns. Is that alright?

Just to be triple-sure.

kathryn-dale (approved):

PR #3107, opened 2026-04-01
BE: Use hardened image

tushortz:
# Description

-

Fixes #CDD-3109

---

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

---

# Checklist:


Dockerfile
```diff
@@ -6,63 +6,59 @@
...
-    # Upgrade the pip installer to the latest version
-    && pip install --no-cache-dir -r requirements-prod.txt
-    # Install project dependencies onto the virtual environment
+# Build runtime bundle and collect shared library deps via dedicated script.
+COPY docker/build_distroless_runtime.sh /usr/local/bin/build_distroless_runtime.sh
+COPY . /code
 
-# Mounts the application code into the image
-COPY . code
+RUN bash /usr/local/bin/build_distroless_runtime.sh
```
mattjreynolds: I think we should consider splitting this out into seperate docker commands for: system packages, python packages, python source code. So something like:
- RUN command for apt-get
- COPY command for requirements fies
- RUN command for installing python dependencies
- COPY command for python source code

This allows us to make use of docker's caching of layers - because each docker command creates a new layer that can be re-used in future builds if none of the underlying files have changed and significantly speeds up build time, especially for the most common dev scenario which is that only python files have been changed.

With how we currently have it, any small change (e.g renaming a variable in a python file) will cause docker to have to install all OS packages, and all python dependencies again, instead of using the cached layers.

Dockerfile
```diff
@@ -6,63 +6,59 @@
 # When bumping Python versions, we currently have to update the `.python-version` file and this `ARG`
 ARG PYTHON_VERSION=3.12.13
 
-FROM python:${PYTHON_VERSION}-slim AS build
+FROM python:${PYTHON_VERSION}-slim-bookworm AS build
```
mattjreynolds, line 10: It might be worth adding a comment for when we come to upgrade the docker images, to show that this debian version needs to match the version of the distroless image below.

docker/build_distroless_runtime.sh
```diff
@@ -0,0 +1,50 @@
...
+    | xargs -r -I '{}' cp -vP --parents '{}' "${DEPS_DIR}/"
+
+  # Copy all shared objects (files AND symlinks) from architecture-specific system lib dirs.
+  for libdir in /lib/*-linux-gnu /usr/lib/*-linux-gnu; do
+    [[ -d "${libdir}" ]] || continue
+    find "${libdir}" \( -type f -o -type l \) -name '*.so*' -print \
+      | xargs -r -I '{}' cp -vP --parents '{}' "${DEPS_DIR}/"
+  done
+
+  apt-get update
```
mattjreynolds: I don't think there's any need for the 2nd `update` here - we already ran `update` on line 17 - so we already have an up-to-date list of packages, and `purge` doesn't use that package list anyway, it just looks at what's already installed

jrdh (approved):

mattjreynolds (approved):

PR #3134, opened 2026-04-15
Task/cdd 3172 permission hierarchy

luketowell:
# Description

- the implementation of permission hierarchy groupings within the api. 
- In this PR, 2 endpoints are introduced which enable a user to retrieve permission sets from the DB and to retrieve grouped permission sets. 
- The Grouped permission sets allow users to retrieve permission sets which adhere to the theme, sub-theme, topic and metric hierarchy e.g. if a user has access to all themes then the permission set returned with be the highest level permission set rather than returning every permission set applied to the user. 
- Functionality for grouping permission set hierarchys by geography_type and by theme is also introduced and documented via the swagger.

Fixes #CDD-3172

---

## Type of change

---

# Checklist:


auth_content/models/users.py
```diff
@@ -12,5 +14,7 @@ class User(models.Model):
         FieldPanel("permission_sets", widget=forms.CheckboxSelectMultiple),
     ]
 
+    objects = UserManager()
```
kathryn-dale, line 17: What's this for?

cms/dynamic_content/blocks.py
```diff
@@ -1,6 +1,15 @@
 from django.core.exceptions import ValidationError
 from django.db import models
-from wagtail import blocks
+from wagtail.blocks import (
```
kathryn-dale, line 3: More an FYI but I expect this will be thrown up by QA as needing a separate PR (I made the same change and had to revert it, just haven't gotten around to the PR yet)

metrics/api/serializers/user.py
```diff
@@ -0,0 +1,302 @@
...
+from metrics.utils.permission_hierarchy import (
+    build_permission_hierarchy,
+    get_deduplicated_permissions,
+)
+
+
+def _validate_user_id(value):
+    """Validate theme_id is either wildcard or a valid integer"""
+    try:
+        uuid_obj = uuid.UUID(value, version=4)  # noqa: F841
```
kathryn-dale: Is the uuid_obj an inbuilt thing?

auth_content/models/users.py
luketowell: It hooks the UserManager up to the Model so that the User class can easily be used to query the DB. its the same pattern that has been applied elsewhere in the code base.

cms/dynamic_content/blocks.py
luketowell: Its the result of a merge conflict from code that is already in the code. This just needs updating. It makes sense for it to be in this PR rather than going through all the effort of creating a new PR merging to main and then rebasing.

metrics/api/serializers/user.py
luketowell: Its just used for validating the UUID that is entered. The stack overflow documentation I found told me to do it that way.

kathryn-dale (approved):

metrics/api/serializers/user.py
```diff
@@ -0,0 +1,336 @@
...
+from auth_content.models.users import User
+from metrics.utils.permission_grouping import group_by_geography_type, group_by_theme
+from metrics.utils.permission_hierarchy import (
+    build_permission_hierarchy,
+    get_deduplicated_permissions,
+)
+
+
+def _validate_user_id(value):
+    """Validate theme_id is either wildcard or a valid integer"""
```
mattjreynolds: Docstring wants updating

metrics/api/urls_construction.py
```diff
@@ -320,7 +321,6 @@ def construct_urlpatterns(
             )
         case enums.AppMode.PRIVATE_API.value:
             constructed_url_patterns += private_api_urlpatterns
-            constructed_url_patterns += permission_set_urlpatterns
         case enums.AppMode.FEEDBACK_API.value:
```
mattjreynolds, line 311: Why did this get removed? It breaks the JWT lambda as we are no longer able to fetch the permission sets.

metrics/api/serializers/user.py
mattjreynolds: Updated

metrics/api/urls_construction.py
mattjreynolds: This is being discussed on the [ticket](https://ukhsa.atlassian.net/browse/CDD-3172) - the plan is to leave this as is so we can merge in this code, and fix it as part of [CDD-3057](https://ukhsa.atlassian.net/browse/CDD-3057) which is where the endpoint will start being consumed by the lambda.

PR #3137, opened 2026-04-17
CDD-3147: Update Cognito User for permission sets

mattjreynolds:
- Add permission sets to request.user object
- Make cognito user ephemeral for speed - DB access is not needed 
- Move auth header name to settings for flexibility

-

Fixes #CDD-3147


metrics/api/settings/default.py
```diff
@@ -115,6 +115,7 @@
 
 COGNITO_USER_MANAGER = "common.auth.cognito_jwt.user_manager.CognitoManager"
 COGNITO_AWS_REGION = config.COGNITO_AWS_REGION
+COGNITO_JWT_AUTH_HEADER = config.COGNITO_JWT_AUTH_HEADER
```
kathryn-dale, line 118: Do we only need this env var in the private api?

kathryn-dale: I think for now it's fine, but do we want to set the default here, rather than at the point of calling the variable?

common/auth/cognito_jwt/backend.py
```diff
@@ -50,6 +51,8 @@ def authenticate(self, request):
         else:
             user_model = self.get_user_model()
             user = user_model.objects.get_or_create_for_cognito(jwt_payload)
+        if not user:
```
kathryn-dale, line 58: Do we want to log anything here to say that there is no user? 🤔 I'm thinking of debugging scenarios, would it be helpful?

mattjreynolds: Added debug logging. I've also updated the logging output in the user manager when fields are missing to improve readability and moved that to be debug level too so we don't spam the logs.

metrics/api/settings/default.py
mattjreynolds: This is in keeping with the other env vars used by this piece of code in that they are optional and only need adding if you are doing something non-standard, and use the Django or Industry standard defaults as their default values.
It's a hang-up from the fact that this came from a standalone django add-on package, so I've tried to keep it logically seperated

kathryn-dale (approved):

README.md
kathryn-dale, line 78: I'd put something in here just to say if you want to run the non-pub version (i.e. pass a valid JWT into requests, you'll have to link to a deployed cognito instance as per the guidance below (link to remote infra bit where you've put the comments)

PR #3193, opened 2026-05-14
Search fixes

aidan:
-

Fixes #CDD-<ISSUE_ID>

---

## Type of change

---

# Checklist:

### DIFF
--- a/metrics/api/settings/test.py
+++ b/metrics/api/settings/test.py
@@ -1,0 +1,22 @@
+import os
+
+from metrics.api.settings import ROOT_LEVEL_BASE_DIR
+
+DATA_UPLOAD_MAX_NUMBER_FIELDS = None
+
+DEBUG = True
+
+DATABASES = {
+    "default": {
+        "TIME_ZONE": "Europe/London",
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": os.path.join(ROOT_LEVEL_BASE_DIR, "test.sqlite3"),
+    },
+    "test": {
+        "TIME_ZONE": "Europe/London",
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": os.path.join(ROOT_LEVEL_BASE_DIR, "test.sqlite3"),
+    },
+}
+
+INTERNAL_IPS = ["127.0.0.1"]
--- a/public_api/version_02/views/search.py
+++ b/public_api/version_02/views/search.py
@@ -1,0 +1,33 @@
+from drf_spectacular.utils import extend_schema
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from wagtail.api.v2.serializers import get_serializer_class
+from wagtail.models import Page
+
+from cms.topic.models import TopicPage
+from public_api.version_02.views.base import PUBLIC_API_TAG
+
+
+@extend_schema(tags=[PUBLIC_API_TAG])
+class SearchView(APIView):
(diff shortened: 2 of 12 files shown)


metrics/api/settings/test.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from metrics.api.settings import ROOT_LEVEL_BASE_DIR
+
+DATA_UPLOAD_MAX_NUMBER_FIELDS = None
+
+DEBUG = True
+
+DATABASES = {
```
tushortz, line 9: since this feields are exact replica, could you create it as a variable then reuse it

public_api/version_02/views/search.py
```diff
@@ -0,0 +1,32 @@
...
+class SearchView(APIView):
+    """This endpoint provides search results and could in the future provide
+    autocomplete suggestions etc.
+
+    """
+
+    def get(self, request: Request):
+        search = request.GET.get("search")
+        limit = int(request.GET.get("limit", 0))
+        fields = request.GET.get("fields", ["title", "slug"])
```
tushortz, line 22: my understanding is that fields would be returned as a string. i would recommend not setting the default as a list but rather splitting the `.get` value before then doing an if check.

atm it seems like we're trying to handle `str` and `list`

public_api/version_02/views/search.py
```diff
@@ -0,0 +1,32 @@
...
+        meta = request.GET.get("meta", ["id"])
```
tushortz, line 23: same as above

tests/integration/public_api/v2/views/test_search.py
```diff
@@ -0,0 +1,154 @@
...
+from cms.topic.models import TopicPage
+from tests.factories.cms.topic import TopicPageFactory
+from tests.factories.cms.page import CommonPageFactory
+
+# @pytest.fixture
+# def rare_topics(db) -> [TopicPage]:
+#     for i in range(2):
+#         TopicPageFactory.create(
+#             path=f"topic-rare-{i}",
+#             depth=1,
```
tushortz: we can remove this if not needed

tests/integration/public_api/v2/views/test_search.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        Given a string that matches 2 topic page
+        When GET /api/public/v2/search is called with that query and no limit
+        Then the results will include the matching topic pages and others with the topics first
+        """
+
+        # Given
+        search = "rare"  # All topics have a title, only even ones have rare in it
+        query = f"search={search}"
+        expected_results = [t for t in topics if "rare" in t.title]
```
tushortz, line 54: would it be best to use Django's `__icontains` filter here instead of a forloop?

bargrap.http
```diff
@@ -0,0 +1,34 @@
+POST http://localhost:8000/api/charts/v3
+Content-Type: application/json
+
```
tushortz: What are the purpose of the `.http` files?

public_api/version_02/views/search.py
```diff
@@ -0,0 +1,32 @@
...
+        topic_results = TopicPage.objects.all().search(search)
+        if not limit or topic_results.count() < limit:
+            # TODO: go get more
```
tushortz, line 26: we may not need the extra `count()` db query for every search. i think you can check if limit is set then do the limit

tushortz (changes requested):

PR #3199, opened 2026-05-18
Cdd 3090 simplified BE chart watermark

kathryn-dale:
# Description

- Add watermark to charts and graphs for non public data

Fixes #CDD-3090

---

## Type of change

---

# Checklist:


mattjreynolds (approved):
👍

mattjreynolds (approved):

metrics/domain/models/charts/common.py
```diff
@@ -20,6 +20,8 @@ class BaseChartRequestParams(BaseModel):
     request: Request | None = None
     confidence_intervals: bool | None = False
     confidence_colour: str | None = ""
+    is_public: bool | None = True
```
dandammann, line 23: Do we want to make the "is_public" querystring and therefore attribute mandatory throughout (front-end passing and back-end receiving)!?

This is in case bad code regressions don't result in "is_public" being passed. Mistakes could creep in, if we allow the parameter to be optional. The decision of whether "is_public" true/false seems too important for it to be optional.

tests/unit/metrics/interfaces/charts/common/test_chart_output.py
```diff
@@ -0,0 +1,207 @@
...
+        # Then
+        assert result == ""
+
+    from unittest import mock
+
+
+class TestPostInitAppliesWatermark:
+    @mock.patch(f"{MODULE_PATH}.ChartOutput._apply_watermark")
+    @mock.patch(f"{MODULE_PATH}.AUTH_ENABLED", True)
+    def test_applies_watermark_when_not_public_and_classified_and_auth_enabled(
```
dandammann, line 19: This function name includes "...when_not_public..." which helps the understanding quite a lot of what is happening here and why.

All the other test functions below could do with the same, i.e. the function name better says whether ChartOutput has "is_public" passed and what value it has.

dandammann: Did you see my comment above?

mattjreynolds (approved):

PR #3201, opened 2026-05-20
CDD-3173: add the base code for the DB permission filtering of privat…

dandammann:
# Description

CDD-3174: Similar to the changes that were implemented for the getPages endpoint as part of https://ukhsa.atlassian.net/browse/CDD-3171, this ticket concerns endpoints that are used for displaying data / generating charts should be updated to make use of the authorisation header. If the user provides an authorization header then the data returned from the endpoint should be customised depending on the permissions that the user has been granted. 

CDD-3173: The database queries will need to be updated for non-public users so that they only return non-public data and they only return data that the user is permitted to view.

When a request is made by an authenticated user, the query to the DB should be made for only the data that the user is able to see. This will be driven from the user permission hierarchy

When a request is made by an authenticated user the returned data should only be for non-public data. it should not contain public data.

---

## Type of change

- [  ] Bug fix (non-breaking change which fixes an issue)
- [  ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [  ] Tech debt item (this is focused solely on addressing any relevant technical debt)

---

# Checklist:

- [  ] I have made corresponding changes to the documentation
- [  ] I have added screenshots or screen grabs where appropriate


common/auth/cognito_jwt/user_manager.py
```diff
@@ -25,7 +45,50 @@ def get_or_create_for_cognito(jwt_payload):
...
+        )
+
         user_class = get_user_model()
         user = user_class(username=username)
         user.permission_sets = permission_sets
+
         return user
+
+
+def extract_jwt_permissions(*, request: "Request | None") -> dict:
```
kathryn-dale, line 64: This feels redundant, why would we call this in every place rather than just looking at `request.user.permission_sets` which is provided to every endpoint anyway? Feels excessive

common/auth/cognito_jwt/backend.py
```diff
@@ -46,7 +51,8 @@ def authenticate(self, request):
         try:
             token_validator = self.get_token_validator(request)
             jwt_payload = token_validator.validate(jwt_token)
-        except TokenError:
+        except TokenError as error:
```
kathryn-dale, line 53: Arguably unneccessary? It's already throwing an error and I'm worried there's a risk that some permission set info or user info could be spat out as part of the error log

metrics/data/managers/core_models/headline.py
```diff
@@ -382,16 +389,40 @@ def query_for_data(
...
+                    jwt_permissions=jwt_permissions,
+                    theme=theme,
+                    sub_theme=sub_theme,
+                    topic=topic,
+                    metric=metric,
+                    geography_type=geography_type,
+                    geography=geography,
+                )
+        else:
+            # THESE LEGACY RBAC PERMISSIONS ARE NOT IN USE AND TO BE REMOVED IN A FUTURE RELEASE
```
kathryn-dale: If they aren't in use in this file can we remove them as part of this work? I don't see any reason to keep them in, as you are essentially replacing what they are trying to do with the new solution :)

metrics/data/managers/core_models/time_series.py
```diff
@@ -611,16 +618,40 @@ def query_for_data(
...
+                    jwt_permissions=jwt_permissions,
+                    theme=theme,
+                    sub_theme=sub_theme,
+                    topic=topic,
+                    metric=metric,
+                    geography_type=geography_type,
+                    geography=geography,
+                )
+        else:
+            # THESE LEGACY RBAC PERMISSIONS ARE NOT IN USE AND TO BE REMOVED IN A FUTURE RELEASE
```
kathryn-dale: Same as above

metrics/utils/permission_hierarchy.py
```diff
@@ -21,6 +21,54 @@
...
+                }
+            ],
+            "summary": {
+                "has_global_access": False
+            },
+        }
+
+    @return {dict} example:
+        {
+            "permission_set_hierarchy": [
```
kathryn-dale, line 45: What's the value in this over using the raw permissions?

metrics/utils/permissions.py
```diff
@@ -0,0 +1,289 @@
...
+                        "sub_theme": {"id": "133", "name": "childhood-vaccines"},
+                        "topic": {"id": "-1", "name": "* (All)"},
+                    }
+                ]
+            }
+
+    @return {bool}
+    """
+
+    topic_record = None
```
kathryn-dale, line 93: Can we not extend the function already created in CDD-3171? I find this hard to read (could just be me though)

common/auth/cognito_jwt/backend.py
mattjreynolds: @dandammann Please can we avoid including work that is not related to the ticket in the same PR.
It makes it harder to review as reviewers have to figure out which code changes are related to the ticket, and which are achieving some other unknown goal.
If you find tech debt that need to be improved In existing code (and it's a quick fix), then these should be added to a seperate PR with a description that explains it and then it can be reviewed on it own merits.
Also in this instance, all these changes within the JWT code are going to cause a big headache of conflicts once the PR for CDD-3147 gets merged in too.

metrics/utils/permission_hierarchy.py
mattjreynolds: @dandammann I''m not sure what this is for. I was under the impression that the permission set hierarchy was already returning the permissions sets into the JWT in the most usable format for the endpoints to consume when filtering the data. 
If the permission sets are not in the right format when we come to use them for filtering for pages, DB data etc, then we should be updating the format of the permission sets before they go into the JWT, so it happens just once, rather than on 100s of requests at the point it is being consumed (although there is also a tradeoff to consider here too, of avoiding making the JWT too large).

common/auth/cognito_jwt/user_manager.py
mattjreynolds: @dandammann  None of the changes in this file are needed either. 
Possibly the bit you are missing that caused this unneccesary work is that `request.user.is_authenticated` will be `True` if it is a non-public request and `False` for a public one. As Kathryn then said, you can just user `request.user.permission_sets` within the endpoint to get the permission_sets hierarchy that was extracted from the JWT.

metrics/api/views/charts/single_category_charts.py
```diff
@@ -218,6 +220,7 @@ def post(cls, request, *args, **kwargs):
 
 
 class EncodedChartsView(APIView):
+    authentication_classes = [SessionAuthentication, JSONWebTokenAuthentication]
```
mattjreynolds, line 223: This should not be being set.
Authenticaion is already being handled for all views by the JWT token code, and the permission sets are being made available in the request object (`request.user.permission_sets`)

metrics/data/managers/core_models/headline.py
```diff
@@ -382,16 +390,38 @@ def query_for_data(
...
-        )
+
+        has_access_to_non_public_data: bool
+
+        if jwt_permissions:
+            # Check JWT permissions first (new authorization takes precedence)
+            has_global_access = jwt_permissions.get("has_global_access", False)
+
+            if has_global_access:
+                has_access_to_non_public_data = True
```
mattjreynolds, line 403: Could simplify this `has_global_access` check to just be part of the return value from `check_any_permissions_allow_access` 
This is how it's being done in 3171, are we not able to just reuse those functions for these endpoints too? They should be doing the same job and we don't want to have 2 different function that both check permissions

metrics/interfaces/plots/access.py
```diff
@@ -162,7 +163,9 @@ def get_queryset_from_core_model_manager(
             plot_params["fields_to_export"].append("lower_confidence")
 
         return self.core_model_manager.query_for_data(
-            **plot_params, rbac_permissions=self.chart_request_params.rbac_permissions
+            **plot_params,
+            rbac_permissions=self.chart_request_params.rbac_permissions,  # legacy permissions to be removed
+            jwt_permissions=extract_jwt_permissions(request=self.chart_request_params.request),  # new permissions
```
mattjreynolds, line 168: as with previous comment, you should be able to just use something like:
`jwt_permissions=request.user.permission_sets if request.auth else None`,

metrics/utils/permissions.py
mattjreynolds: I agree, we want to make sure we're not duplicating the core permissions checking functions

dandammann: Hi @kathryn-dale and @mattjreynolds . Please look at my comments and examples in this current function and filter_permissions() and let me know, if you think your check_permissions() can achieve that.

I myself think it is too simplistic for that, because it only checks permissions for EITHER theme, sub_theme or topic, but with the charts we need to check permissions cross-cutting (taking them all into consideration).

if you do think check_permissions() is adequate, please try it out ...

dandammann: There is also that difference that with GetPages the input arguments are theme/sub_theme/topic IDS, whereas with charts they are theme/sub_theme/topic NAMES.

metrics/data/managers/core_models/headline.py
dandammann: I mentioned that RBAC permissions need removing (at some point), but you said that's out-of-scope, because there is a ticket for that in the backlog.

metrics/data/managers/core_models/time_series.py
dandammann: Same reply from me too.

metrics/data/managers/core_models/headline.py
mattjreynolds: For the areas of the code you are already touching, it's fine to replace the RBAC parts with the new permission_sets bits, rather than adding if/else statements for code we know is going to disappear soon.
The endpoints are going to be thoroughly tested by updated unit tests and by the QA team anyway, so there's no issues with doing it that way, it's just that we don't need to make more work by extending the scope of the ticket to cover removing all the other areas of the codebase that have RBAC bits, along with all the additional updated tests and additional QA time that would go along with that work. That part can be done a future ticket.
We want to focus on what's required for our MVP deliverable, and keeping that work as simple as possible.

common/auth/cognito_jwt/user_manager.py
dandammann: Done.

request.user.permission_sets didn't work for me at the time, but that might have been cause I called it in the wrong place. It works now.

PR #3212, opened 2026-05-27
CDD-3351: Audit logging

itsthatianguy:
# Description

- Adds new audit logger to the application
- New middleware to get the current user so the user making the change in the CMS can be logged
- Signals added to monitor for changes to the User, PermissionSet, and the relationship between the two
- Audit logs created when relevant changes are made

---

## Type of change

---

# Checklist:


common/signals.py
```diff
@@ -0,0 +1,128 @@
+import logging
+
+from django.db.models.signals import m2m_changed, post_delete, post_save, pre_save
+from django.dispatch import receiver
+
+from metrics.api.middleware.current_user import get_current_user
+
+audit_logger = logging.getLogger("audit")
+
+AUDITABLE_MODELS = ["PermissionSet", "User"]
```
kathryn-dale, line 10: Will this potentially need updating for the API when we go live with that? I think it might be worth creating a `non public logging guide` in the docs folder to add some notes on when this needs extending 🤔

kathryn-dale (commented):
LGTM but I think that doc would be useful :)

common/signals.py
itsthatianguy: I've added a doc - let me know if it doesn't cover everything we'd expect
But it just needs extending when there are other models we want to audit changes on really

kathryn-dale (approved):

PR #3221, opened 2026-05-29
CDD 3174 - Downloads v2 endpoint

dandammann:
# Description

See https://ukhsa.atlassian.net/browse/CDD-3174

---

## Type of change

- [  ] Bug fix (non-breaking change which fixes an issue)
- [  ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [  ] Tech debt item (this is focused solely on addressing any relevant technical debt)

---

# Checklist:

- [  ] I have made corresponding changes to the documentation
- [  ] I have added screenshots or screen grabs where appropriate


cms/home/migrations/0036_alter_landingpage_make_health_topic_optional.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 5.2.14 on 2026-05-28 16:04
```
kathryn-dale, line 1: As a rule, we don't include migrations unrelated to the ticket in PRs - if you find a migration needs doing and hasn't been merged to main, shout in the team channel to see who owns it and ask them to raise a separate PR for their missing migration. 

This needs to be deleted and main merged in. This fix was done by Josh and Phill and the migration file already merged to main 👍

cms/home/models/landing_page.py
```diff
@@ -41,6 +41,7 @@ class LandingPage(UKHSAPage):
         max_num=HEALTH_TOPIC_SECTION_LINK_MAX_NUM,
         default=list,
         use_json_field=True,
+        blank=True,
```
kathryn-dale, line 44: Same here, should be removed and done when main is merged in

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+    topic_id,
+    metric_id=None,
+    geography_type=None,
+    geography_id=None,
+) -> bool:
+    """
+    Core permission evaluation shared by page and chart permission checks.
+    """
+
+    logger.info("Entered check_permissions()")
```
kathryn-dale: This feels like a debug log to me (if we even need it at all?)

metrics/api/middleware/sql_debug.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def _print_sql(execute, sql, params, many, context):
+    print(f"\n[SQL] {sql}")
+    if params:
+        print(f"[PARAMS] {params}")
+    return execute(sql, params, many, context)
+
+
+class SQLDebugMiddleware:
```
kathryn-dale: This rings PII data breach potential alarm bells to me. How do we ensure it can only be used locally and not accidentally released to production?

metrics/data/managers/core_models/geography.py
```diff
@@ -47,6 +47,19 @@ def get_name_by_code(self, geography_code: str) -> str | None:
             .first()
         )
 
+    def get_id_by_name(self, geography_name: str) -> int | None:
```
kathryn-dale: Nitpick: Can we update the name to include geography? Just to be super clear

metrics/data/managers/core_models/geography.py
```diff
@@ -167,6 +180,23 @@ def get_name_by_code(self, geography_code: int) -> str | None:
         """
         return self.get_queryset().get_name_by_code(geography_code)
 
+    def get_id_by_name(self, geography_name: str) -> int | None:
```
kathryn-dale: What's the difference between this and the one above? They have the same name.

metrics/data/managers/core_models/geography_type.py
```diff
@@ -41,6 +41,19 @@ def get_name_by_id(self, geography_type_id: int) -> str | None:
         """
         return self.filter(id=geography_type_id).values_list("name", flat=True).first()
 
+    def get_id_by_name(self, geography_type_name: str) -> int | None:
```
kathryn-dale, line 44: Same nitpick here, can we be super clear in function naming?

metrics/data/managers/core_models/time_series.py
```diff
@@ -171,8 +174,11 @@ def query_for_data(
         stratum: str | None = None,
         sex: str | None = None,
         age: str | None = None,
+        theme: str = "",
+        sub_theme: str = "",
         metric_value_ranges: list[tuple[str | float | int]] | None = None,
-        restrict_to_public: bool = True,
+        rbac_permissions: Iterable[RBACPermission] | None = None,
```
kathryn-dale: Do we use rbac_permissions? I thought that was the old one that is not in use?

metrics/domain/models/charts/common.py
```diff
@@ -25,5 +28,20 @@ class Config:
...
+        request_user = getattr(self.request, "user", None)
+        return getattr(request_user, "permission_sets", {})
+
+    @property
+    def rbac_permissions(
+        self,
+    ) -> Iterable[
+        object
+    ]:  # set to object instead of RBACPermission, cos of architectural constraints
+        """TODO: RBAC-based permissions are legacy and will be removed in a future release"""
```
kathryn-dale: Remove them here, as you are replacing them with the new stuff. The future ticket is to clean up anything that hasn't been replaced/picked up, but no reason at all why we should leave this dead code here :)

metrics/domain/models/charts/subplot_charts.py
```diff
@@ -25,6 +27,15 @@ class Config:
     def rbac_permissions(self) -> Iterable["RBACPermission"]:
         return getattr(self.request, "rbac_permissions", [])
 
+    @property
+    def permission_sets(self) -> dict:
+        """Extract optional JWT permissions from the authenticated request"""
+
+        logger.info("Entered subplot_charts.py:BaseChartRequestParams.permission_sets")
```
kathryn-dale: Again, debug log? Or is it required once this is fully tested?

metrics/domain/weather_health_alerts/text_lookups/heat_alert_text.py
```diff
@@ -40,7 +40,7 @@
 </ul>
 """
 _LEVEL_10_TEXT = """
-<p>Significant impacts are <strong>possible</strong> across the health and social care services due to the high temperatures, including:</p>
+<p>Significant impacts are <strong>possible</strong> across health and social care services due to the high temperatures, including:</p>
```
kathryn-dale, line 43: Why has this changed? Do the designs have the different wording?

metrics/domain/weather_health_alerts/text_lookups/heat_alert_text.py
```diff
@@ -51,7 +51,7 @@
 </ul>
 """
 _LEVEL_11_TEXT = """
-<p>There is potential for severe impacts across health and social care services due to the high temperatures, including:</p>
+<p>There is <strong>potential</strong> for severe impacts across health and social care services due to the high temperatures, including:</p>
```
kathryn-dale, line 54: Why has this formatting changed? Is this due to the designs? I don't understand how it's relevant for this ticket

metrics/interfaces/plots/access.py
```diff
@@ -147,6 +147,8 @@ def get_queryset_from_core_model_manager(
                     ('05-10', Decimal('9.0'))
                 ]>`
         """
+        logger.info("Entered access.py:get_queryset_from_core_model_manager()")
```
kathryn-dale: Debug log again?

metrics/interfaces/plots/access.py
```diff
@@ -162,7 +164,9 @@ def get_queryset_from_core_model_manager(
             plot_params["fields_to_export"].append("lower_confidence")
 
         return self.core_model_manager.query_for_data(
-            **plot_params, rbac_permissions=self.chart_request_params.rbac_permissions
+            **plot_params,
+            rbac_permissions=self.chart_request_params.rbac_permissions,  # old permissions (remove)
```
kathryn-dale, line 166: Can we remove this as part of the PR? We are actively replacing it

requirements-prod.txt
```diff
@@ -3,7 +3,7 @@ anyascii==0.3.3
 attrs==26.1.0
 beautifulsoup4==4.14.3
 CacheControl==0.14.4
-certifi==2026.4.22
```
kathryn-dale: Why are we updating these requirements as part of this ticket? Don't disagree that updating them could be valuable, but they should be their own PR, so that if things break we can see if it's the package or your changes

kathryn-dale (changes requested):

cms/home/migrations/0036_alter_landingpage_make_health_topic_optional.py
dandammann: This change doesn't come from me. I have now merged the up-to-date main into my base ian branch, and ian into mine, and this particular DIFF is gone now.

cms/home/models/landing_page.py
dandammann: This change doesn't come from me. I have now merged the up-to-date main into my base ian branch, and ian into mine, and this particular DIFF is gone now.

metrics/domain/weather_health_alerts/text_lookups/heat_alert_text.py
dandammann: This change doesn't come from me. I have now merged the up-to-date main into my base ian branch, and ian into mine, and this particular DIFF is gone now.

requirements-prod.txt
dandammann: This change doesn't come from me. I have now merged the up-to-date main into my base ian branch, and ian into mine, and this particular DIFF is gone now.

metrics/interfaces/plots/access.py
dandammann: I just tried commenting out this line, and see how much will break. 

Quite a lot, actually.

If you want me to remove this line, a lot more code will have to change as a consequence.

In order to keep this PR reviewable, would you agree we better postpone this removal and do as part of the dedicated RBAC removal ticket in the backlog?

common/auth/permissions.py
dandammann: Please have a look at my ticket comment where I explain what this is and why:

"added logs at function entry points & printed out all SQL for debugging purposes (whilst ticket 3174 is still in flow)"

This is temporary and equivalent to your print() statements in your CDD-3171 ticket.

metrics/api/middleware/sql_debug.py
dandammann: Please have a look at my ticket comment where I explain what this is and why:

"added logs at function entry points & printed out all SQL for debugging purposes (whilst ticket 3174 is still in flow)"

This SQL has helped Matt and me to solve a bug during our pair-programming, so we hope it will come handy for our testers too, for whom I wrote the JIRA comment.

In terms of risk, the way how I understand it that this current CI failure will NOT allow this code to accidentally end up in production:

error[T201]: `print` found
 --> metrics/api/middleware/sql_debug.py:5:5
  |
4 | def _print_sql(execute, sql, params, many, context):
5 |     print(f"\n[SQL] {sql}")

metrics/data/managers/core_models/geography.py
dandammann: One method is from the GeographyQuerySet and the other from the GeographyManager class (the Django way of doing things).

The naming convention so far is that we call them the same name, so I continued doing that.

metrics/data/managers/core_models/geography_type.py
dandammann: Whilst we are in the GeographyTypeQuerySet and GeographyManager classes of the geography_type.py most methods don't repeat "geography_type" again the method names, eg:

get_name_by_code()
get_all_names()
get_name_by_id()

My get_id_by_name() follows on from that naming convention.

When accessing this method from outside the class the context also becoems clear:

geography_type_model: Any = apps.get_model("data", "GeographyType")
geography_type_id = geography_type_model.objects.get_id_by_name(geography_type)

metrics/data/managers/core_models/geography.py
dandammann: See my reply below.

metrics/data/managers/core_models/time_series.py
dandammann: This I could remove quite easy without any other failures, so I removed it now.

metrics/domain/models/charts/common.py
dandammann: I just tried commenting out this method, and see how much will break.

Quite a lot, actually.

If you want me to remove this line, a lot more code will have to change as a consequence.

In order to keep this PR reviewable, would you agree we better postpone this removal and do as part of the dedicated RBAC removal ticket in the backlog?

metrics/domain/models/charts/subplot_charts.py
dandammann: See my reply above.

cms/home/models/landing_page.py
kathryn-dale: I know it doesn't come from you, that's why I called it out :)

metrics/domain/models/charts/common.py
kathryn-dale: Yep that makes sense 👍

metrics/data/managers/core_models/time_series.py
```diff
@@ -231,6 +240,9 @@ def query_for_data(
                     ]>`
 
         """
+
+        logger.info("Entered query_for_data()")
```
kathryn-dale: Do we need this log? Not sure how useful it is after this work has been completed

metrics/data/managers/core_models/time_series.py
```diff
@@ -245,9 +257,30 @@ def query_for_data(
             sex=sex,
             age=age,
         )
-
-        if restrict_to_public:
-            queryset = queryset.filter(is_public=True)
+        public_queryset = queryset.filter(is_public=True)
+
+        if permission_sets:
+            logger.info("Entered if permission_sets clause")
```
kathryn-dale: I'd remove this log before QA too

metrics/data/managers/core_models/time_series.py
```diff
@@ -245,9 +257,30 @@ def query_for_data(
...
+                geography_name=geography,
+            ):
+                logger.info("Entered check_permissions_by_name() if clause")
+
+                # Spec says: Deliver both public and non-public data
+                queryset = public_queryset | queryset.filter(is_public=False)
+            else:
+                queryset = public_queryset
+        else:
+            logger.info("Entered else permission_sets clause")
```
kathryn-dale: Same here, this feels useful for dev work but too noisy when we go live with it

metrics/data/managers/core_models/topic.py
```diff
@@ -31,15 +31,38 @@ def get_name_by_id(self, topic_id: int) -> str | None:
 
         Returns:
             The topic name if found, None otherwise
-
-        Examples:
```
kathryn-dale, line 35: Why has this been removed? It seems helpful

metrics/domain/models/charts/common.py
```diff
@@ -25,5 +28,20 @@ class Config:
...
+
+        request_user = getattr(self.request, "user", None)
+        return getattr(request_user, "permission_sets", {})
+
+    @property
+    def rbac_permissions(
+        self,
+    ) -> Iterable[
+        object
+    ]:  # set to object instead of RBACPermission, cos of architectural constraints
```
kathryn-dale: Can we formalize this comment, and explain which architectural constrain this is fulfilling?

metrics/domain/models/charts/common.py
```diff
@@ -25,5 +28,20 @@ class Config:
         arbitrary_types_allowed = True
 
     @property
-    def rbac_permissions(self) -> Iterable["RBACPermission"]:
+    def permission_sets(self) -> dict:
+        """Extract optional JWT permissions from the authenticated request"""
+
+        logger.info("Entered common.py:BaseChartRequestParams.permission_sets")
```
kathryn-dale: This needs removing before QA to minimize log noise

metrics/interfaces/plots/access.py
kathryn-dale: Yep that's fine, I didn't think it would be so large

common/auth/permissions.py
kathryn-dale: The print statements shouldn't still be in the PR, I thought I'd removed them but if not that's a mistake on my part. I wouldn't have anything in the PR that we don't plan to release, as that is a code change going in after it's been reviewed

metrics/domain/models/charts/subplot_charts.py
dandammann: I have now removed every logger.info("Entered ...).

metrics/interfaces/plots/access.py
dandammann: I have now removed every logger.info("Entered ...).

metrics/data/managers/core_models/time_series.py
dandammann: I have now removed every logger.info("Entered ...).

metrics/data/managers/core_models/time_series.py
dandammann: I have now removed every logger.info("Entered ...).

metrics/data/managers/core_models/time_series.py
dandammann: I have now removed every logger.info("Entered ...).

metrics/domain/models/charts/common.py
dandammann: I have now removed every logger.info("Entered ...).

metrics/domain/models/charts/common.py
dandammann: I reverted it back to what it was now, so it doesn't need a comment at all.

We'll have an extra warning for now, but rbac_permissions() soon gets removed anyway.

metrics/data/managers/core_models/topic.py
dandammann: Quite a long noisy function header I thought for such a simple function, but I re-added that comment now anyway.

metrics/domain/weather_health_alerts/text_lookups/heat_alert_text.py
dandammann: This change doesn't come from me. I have now merged the up-to-date main into my base ian branch, and ian into mine, and this particular DIFF is gone now.

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
+import logging
+from typing import Any
+
+from django.apps import apps
+
+WILDCARD_ID_VALUE = "-1"
```
jrdh, line 5: We should only have one version of this constant I think (the existing one is in `cms/auth_content/constants.py`). Given the architecture boundaries, I'd imagine using the new one you've added here in `common` is better than using the `cms/auth_content` one? Either way though, I don't think we should have `WILDCARD_ID_VALUE = "-1"` defined in two separate python modules.

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+
+def check_permissions_by_name(
+    *,
+    permission_sets,
+    theme_name,
+    sub_theme_name,
+    topic_name,
+    metric_name,
+    geography_type,
+    geography_name,
```
jrdh: Would be nice to add type annotations to these.

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+        return False
+    if not isinstance(permission_sets.get("permission_sets"), list):
+        return False
+    if not isinstance(permission_sets.get("summary"), dict):
+        return False
+    if not isinstance(permission_sets.get("summary").get("has_global_access"), bool):
+        return False
+
+    if permission_sets.get("summary").get("has_global_access"):
+        return True
```
jrdh, line 54: I started writing a comment about changing the set of checks here into something shorter and more concise but then decided that actually I think we should wrap the permission sets dict we extract from the JWT into an object so that we can associate all this logic with the permission sets instead of copying it around anywhere we want to use the permission sets. It's cleaner in my opinion.

So my suggestion would be that we create a class (in this module probably, or a new one?) which looks something like this:

```python
# custom error for mistyped permission sets, could be handled differently, not fussed
class PermissionSetsTypeError(Exception):
    def __init__(self):
        super().__init__("Permission sets should be a dict")

@dataclass
class UserPermissions:
    permission_sets: list[dict]
    summary: dict | None

    @property
    def has_global_access(self) -> bool:
        if not self.summary:
            return False
        has_global_access = self.summary.get("has_global_access", False)
        return isinstance(has_global_access, bool) and has_global_access

    def __iter__(self) -> Generator[dict]:
        if not isinstance(self.permission_sets, list):
            return
        yield from self.permission_sets

    @classmethod
    def create(cls, permission_sets: dict) -> Self:
        if not isinstance(permission_sets, dict):
            raise PermissionSetsTypeError
        return cls(permission_sets=permission_sets.get("permission_sets", []), summary=permission_sets.get("summary"))

```

I'd suggest we instantiate this at the entry point in the `CognitoManager.get_or_create_for_cognito` function and change `user.permission_sets = permission_sets` to `user.permission_sets = UserPermissions.create(permission_sets)`. Then whenever we use `user.permission_sets` we have a safe type annotated object with handy properties to access the values in the permission sets structure.

Happy to discuss this!

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+    metric_model: Any = apps.get_model("data", "Metric")
+    geography_type_model: Any = apps.get_model("data", "GeographyType")
+    geography_model: Any = apps.get_model("data", "Geography")
+
+    theme_id, sub_theme_id, topic_id = topic_model.objects.get_id_by_name(
+        theme_name, sub_theme_name, topic_name
+    )
+    metric_id = metric_model.objects.get_id_by_name(metric_name)
+    geography_type_id = geography_type_model.objects.get_id_by_name(geography_type)
+    geography_id = geography_model.objects.get_id_by_name(geography_name)
```
jrdh: I think this violates the architecture constraints it just does it in a way that the checker can't detect. Should we not be using the interfaces for this (e.g. `MetricsAPIInterface`)?

_(As an aside, I like the existance of the architectural checks and the interface pattern, but I think it could be written differently and it really makes the code quite messy... but that is a future problem not a now problem!_

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+
+def check_permissions(
+    permission_sets,
+    theme_id,
+    sub_theme_id,
+    topic_id,
+    metric_id=None,
+    geography_type=None,
+    geography_id=None,
+) -> bool:
```
jrdh, line 103: Again, type annotations would be nice here.

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+        elif check_metric_related_permissions(
+            permission_set=permission_set,
+            theme_id=theme_id,
+            sub_theme_id=sub_theme_id,
+            topic_id=topic_id,
+            metric_id=metric_id,
+        ):
+            return True
+
+    return False
```
jrdh, line 184: Given there's a duplicate if here, assuming that is intended, I think this could be simplified:

```python
for permission_set in permission_sets:
    if check_metric_related_permissions(
        permission_set=permission_set,
        theme_id=theme_id,
        sub_theme_id=sub_theme_id,
        topic_id=topic_id,
        metric_id=metric_id,
    ):
        if geography_type and geography_id and not check_geography_permissions( # noqa: SIM103
            permission_set=permission_set,
            geography_type=geography_type,
            geography_id=geography_id,
        ):
            return False
        return True

return False

```

Pretty sure that logic is identical, but haven't run it through the unit tests to check so will need validating.

Alternate option if we want to keep the clear `return True` driven logic (not unreasonable!):

```python
for permission_set in permission_sets:
    base_check = check_metric_related_permissions(
        permission_set=permission_set,
        theme_id=theme_id,
        sub_theme_id=sub_theme_id,
        topic_id=topic_id,
        metric_id=metric_id,
    )
    if geography_type and geography_id:
        if base_check and check_geography_permissions(
            permission_set=permission_set,
            geography_type=geography_type,
            geography_id=geography_id,
        ):
            return True
    elif base_check:
        return True

return False
```

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def check_metric_related_permissions(
+    *,
+    permission_set,
+    theme_id,
+    sub_theme_id,
+    topic_id,
+    metric_id=None,
+) -> bool:
```
jrdh, line 193: Type annotations again 🙂

common/auth/permissions.py
```diff
@@ -0,0 +1,204 @@
...
+
+    return False
+
+
+def check_geography_permissions(
+    *,
+    permission_set,
+    geography_type=None,
+    geography_id=None,
+) -> bool:
```
jrdh, line 273: Type annotations again 🙂

metrics/api/middleware/sql_debug.py
jrdh: Surely if we've written enough tests, we shouldn't need to rely on manual checks for correctness? Completely understand it from the point of view of while you're developing the feature, but I wouldn't expect QAT to need this as they'll be concentrating on whether the user's experience matches expectations. Whether the code works as expected (e.g. sends the right SQL to the database) is on us, not them surely?

metrics/domain/models/charts/subplot_charts.py
```diff
@@ -25,6 +27,15 @@ class Config:
...
+
+        request_user = getattr(self.request, "user", None)
+        return getattr(request_user, "permission_sets", {})
```
jrdh: I don't have an issue with this pattern of having a helper to get hold of the user's permissions sets that were extracted from the JWT, however, I think it should be moved out of the `Subplots` class and made a separate, shared function then we can use it everywhere we currently have `req.user.permission_sets` or similar (and if we do [the suggestion in that comment above](https://github.com/UKHSA-Internal/data-dashboard-api/pull/3221/changes#r3343162169) we can have a nice typed return too.

jrdh (changes requested):
Made some comments throughout, hopefully they're useful and make sense but happy to discuss 🙂 
Additionally, I think this needs lots more tests. That might have been already on your radar, but I would have expected them to go in with the initial versions of the new checking functions before the code review took place tbh - unless that review was very clearly stated as a WIP, early access check this is ok review, which this kinda half is and isn't. Either way, lots of tests for all the scenarios that could happen with permission sets is warranted here as it's the critical logic path protecting the data.

metrics/api/middleware/sql_debug.py
kathryn-dale: Agreed, while I see the value during development, I don't think this should go in as part of the commit

dandammann: No worries, I've removed it now.

common/auth/permissions.py
dandammann: Done.

There is a bigger issue here, which is that our permission resource ids (eg theme_id, etc ...) are being passed around as strings in the python and then saved as varchar() in the DB. That makes it impossible then to have key constraints to the integer data_theme.id column (same with data_sub_theme and the other master tables).

Such key constraints would prevent empty "" permissions being inserted into the DB, as we experienced through testing.

Basically, what I am suggesting is that not only the APP, but also the DB defensively applies type-checks and best-practice safeguards that are common place in the industry.

Especially, when permissions are concerned. Agreed?

common/auth/permissions.py
dandammann: Again, a good idea.

Matt will trial that in his own CDD-3174-OO (or similar) branch whilst I get your other suggestions ready for QA ... so we don't miss the boat.

common/auth/permissions.py
dandammann: Done.

common/auth/permissions.py
dandammann: Done.

common/auth/permissions.py
dandammann: Done.

common/auth/permissions.py
dandammann: Done.

common/auth/permissions.py
dandammann: Done.

common/auth/permissions.py
dandammann: Done.

metrics/domain/models/charts/subplot_charts.py
dandammann: Done.

I've added a BaseRequestParams() class now that all other request classes (directly or indirectly) inherit.

common/auth/permissions.py
kathryn-dale: Can Matt's stuff now be brought in here? (I think it's done?), or is it better to review separately?

common/auth/permissions.py
```diff
@@ -0,0 +1,358 @@
...
+        )
+
+        # All permission fields must be present
+        if permission_ids is None:
+            return False
+        (
+            permission_theme_id,
+            permission_sub_theme_id,
+            permission_topic_id,
+            permission_metric_id,
```
kathryn-dale, line 149: So in getPages for example, metric won't be provided. I can see pages are handled below but are there any other cases where one of the above might legitimately not be provided? Maybe not one for this PR but something to think about and update if necessary going forward

metrics/data/managers/core_models/time_series.py
```diff
@@ -246,7 +253,19 @@ def query_for_data(
...
+            if not check_chart_permissions_by_name(
+                permission_sets=permission_sets,
+                theme_name=theme,
+                sub_theme_name=sub_theme,
+                topic_name=topic,
+                metric_name=metric,
+                geography_type=geography_type,
+                geography_name=geography,
+            ):
+                queryset = queryset.filter(is_public=True)
```
kathryn-dale: Both branches of this if statement filter the query based on is_public=True. If I'm reading this right: 

IF there are permission sets, BUT the user doesn't have permission, only return public data
ELSE IF there are no permission sets, only return public data

I can't see the branch for returning non public data (i.e. IF there are permission sets, AND the user has permissions, filter and return public + authorized non public data)

metrics/domain/models/charts/common.py
```diff
@@ -1,12 +1,10 @@
...
 from decimal import Decimal
 from typing import Literal
 
-from pydantic.main import BaseModel
-from rest_framework.request import Request
+from metrics.domain.models.common import BaseRequestParams
 
 
-class BaseChartRequestParams(BaseModel):
+class ChartRequestParams(BaseRequestParams):
```
kathryn-dale: How come this has been renamed?

common/auth/logging.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if isinstance(permission_sets, dict):
+        summary = permission_sets.get("summary", {})
+        total_permission_sets = summary.get("total_permission_sets", "unknown")
+        has_global_access = summary.get("has_global_access", False)
+    elif isinstance(permission_sets, list | tuple):
+        if not permission_sets:
+            return
+
+        total_permission_sets = len(permission_sets)
```
kathryn-dale: This is provided on the summary, so can just use that directly rather than recalculate

common/auth/logging.py
```diff
@@ -0,0 +1,52 @@
...
+        has_global_access = any(
```
kathryn-dale: We are already given this information in the summary, what's the value in recalculating it?

metrics/domain/models/plots.py
```diff
@@ -141,7 +141,7 @@ def line_colour_enum(self) -> RGBAChartLineColours:
         return RGBAChartLineColours.BLACK
 
 
-class ChartRequestParams(BaseChartRequestParams):
+class ChartRequestParams(ChartRequestParams):
```
mattjreynolds: I really don't like this, and it will cause confusion in the future.
If a second class lower down in this file inherits from `ChartRequestParams`, it's going to inherit from the newly defined one, not from the one that was imported from `common`, which is probably not the intended behaviour.

tests/integration/metrics/data/managers/core_models/test_time_series.py
```diff
@@ -255,19 +252,90 @@ def test_query_for_data_includes_non_public_records_when_restrict_to_public_is_f
...
+                        )
+                    },
+                    "geography_type": {
+                        "id": str(public_record.geography.geography_type.id)
+                    },
+                    "geography": {"id": str(public_record.geography.id)},
+                }
+            ],
+            "summary": {"has_global_access": False},
+        }
```
mattjreynolds, line 325: This is all built from the public record not the non-public one, which may because that's what the rbac tests used to do, and it won't actually make a difference because the Factory will have used the same data for the public record and the non-public record. I think it's still worth changing as it would be safer and clearer to explicitly use the non-public data for making the permission set.

common/auth/permissions.py
dandammann: Yes, agree. We may have to separate out the current 3-resource permission model (pages) and the 6-resource model (charts) again as we go forward. How is not entirely clear yet, but it will become clear.

metrics/data/managers/core_models/time_series.py
dandammann: I wrote a comment that there isn't actually a DB filter() needed for that condition. A filter() would only complicate the SQL WHERE clause.

I made this comment more prominent now, so it doesn't get missed.

metrics/domain/models/charts/common.py
dandammann: When I introduced my new BaseRequestParams class the old BaseChartRequestParams name became confusing:

- the word Base in BaseChartRequestParams can make you confuse it with BaseRequestParams
- BaseChartRequestParams has the word Chart in the wrong place, so is not pseudo-namespaced against BaseRequestParams anyway
- within its new context it makes more sense to shorten BaseChartRequestParams to ChartRequestParams. It's easier to read short than long class names.

common/auth/logging.py
dandammann: Done.

common/auth/logging.py
dandammann: Done.

metrics/domain/models/plots.py
jrdh: Related to [Kathryn's comment](https://github.com/UKHSA-Internal/data-dashboard-api/pull/3221#discussion_r3375016193), this isn't great - we're redeclaring `ChartRequestParams` which is just confusing?

metrics/domain/models/charts/common.py
jrdh: I think the `Base` prefix is useful here, not least because of this: https://github.com/UKHSA-Internal/data-dashboard-api/pull/3221/changes#r3386687000.

metrics/domain/models/common.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @property
+    def permission_sets(self) -> PermissionSetsType:
+        """Extract optional JWT permissions from the authenticated request"""
+
+        request_user = getattr(self.request, "user", None)
+        return getattr(request_user, "permission_sets", {})
+
+    @property
+    def rbac_permissions(self) -> Iterable["RBACPermission"]:
```
jrdh, line 23: This `RBACPermission` is referenced but never imported.

jrdh: Kinda stunned this didn't get picked up by one of our linting checks tbh. Might look into that later...

common/auth/cognito_jwt/backend.py
```diff
@@ -50,16 +52,22 @@ def authenticate(self, request):
...
             user = user_model.objects.get_or_create_for_cognito(jwt_payload)
+
         if not user:
             logger.debug(
                 "Unable to create user from JWT, defaulting to unauthenticated"
             )
             return None
+
+        # Log user permissions for every JWT API call
+        log_user_permissions(user)
```
mattjreynolds: I think we'd decided we weren't going to log user permissions on every api call because of the increase in logging this would generate, and we were just going to rely on the audit logging piece to ensure we had a record of which permissions a user had at any give time.
@jrdh / @kathryn-dale Can you confirm if I'm remembering this correctly?

common/auth/permissions.py
```diff
@@ -0,0 +1,358 @@
...
+        return False
+
+    if permission_sets.get("summary").get("has_global_access"):
+        return True
+
+    # Sanity check, because front-end will never send empty "" requests
+    if not theme_name or not sub_theme_name or not topic_name or not metric_name:
+        return False
+    if not geography_type or not geography_name:
+        return False
```
mattjreynolds: Do we need both this check and the `if any( value is None` check below? I think the code would be cleaner with just the second check. If we do need both, can we do this in the same style as the check below i.e `if any(not value for value in (...`

common/auth/logging.py
dandammann: Done.

common/auth/logging.py
dandammann: Done.

metrics/data/managers/core_models/time_series.py
```diff
@@ -246,7 +253,21 @@ def query_for_data(
...
+                topic_name=topic,
+                metric_name=metric,
+                geography_type=geography_type,
+                geography_name=geography,
+            ):
+                # Keep both the public and non-public data
+                pass
+            else:
+                queryset = queryset.filter(is_public=True)
+        else:
```
mattjreynolds: Can we simplify this to:

```
        if permission_sets and check_chart_permissions_by_name(
                permission_sets=permission_sets,
                theme_name=theme,
                sub_theme_name=sub_theme,
                topic_name=topic,
                metric_name=metric,
                geography_type=geography_type,
                geography_name=geography,
            ):
                # Keep both the public and non-public data
                pass
            else:
                queryset = queryset.filter(is_public=True)
```

tests/unit/common/auth/test_permissions.py
```diff
@@ -0,0 +1,1245 @@
...
+                "40",
+                "50",
+                "60",
+                id=("test_sub_theme_wildcard_with_wildcards_following_grants_access"),
+            ),
+            pytest.param(
+                [
+                    {
+                        "theme": {"id": "10"},
+                        "sub_theme": {"id": "-1"},
```
mattjreynolds: sub_theme should have a value other than -1 here

metrics/domain/models/common.py
dandammann: Yes, better check the CI-linter.

My IDE warned me and I changed the RBACPermission type to simply object. Kathryn wanted me to comment on that change, but since I shouldn't have changed this PR-unrelated code anyway, I reverted back to RBACPermission. I now suggest we'll leave it reverted.

tests/unit/common/auth/test_permissions.py
```diff
@@ -0,0 +1,1245 @@
...
+                    }
+                ],
+                "10",
+                "20",
+                "30",
+                "40",
+                "50",
+                "60",
+                id=("test_geography_type_mismatch_with_geography_match_denies_access"),
+            ),
```
mattjreynolds: I think the names for these tests don't match the data. Also I don't think they test anything different to the previous 2 test cases? All 4 seem to cover 'geography_type doesn't match', and 'geography doesn't match'.

tests/unit/common/auth/test_permissions.py
```diff
@@ -0,0 +1,1245 @@
...
+    def test_check_chart_permissions_with_missing_values(
+        self,
+        permission_sets,
+        theme_id,
+        sub_theme_id,
+        topic_id,
+        metric_id,
+        geography_type,
+        geography_id,
+    ):
```
mattjreynolds, line 975: If you add docstrings to these test functions as per `TestCheckPagePermissions` that should be enough to stop sonarcloud complaining the functions are identical.

metrics/domain/models/plots.py
dandammann: That's definitely on oversight of mine.

The most logical line of code here, in my opinion, would be:

class PlotRequestParams(ChartRequestParams):

Let me know, if you agree. I have a working commit of this locally, but it involved a lot of re-importing/re-naming throughout the codebase, which I guess we may want to avoid right now.

So, instead I now went with what I thought to be the next-best solution, which is:

class ChartRequestParams(ChartBaseRequestParams):

dandammann: That's definitely on oversight of mine.

The most logical line of code here, in my opinion, would be:

class PlotRequestParams(ChartRequestParams):

Let me know, if you agree. I have a working commit of this locally, but it involved a lot of re-importing/re-naming throughout the codebase, which I guess we may want to avoid right now.

So, instead I now went with what I thought to be the next-best solution, which is:

class ChartRequestParams(ChartBaseRequestParams):

metrics/domain/models/charts/common.py
dandammann: Done.

common/auth/cognito_jwt/backend.py
dandammann: The audit logging only addresses CMS actions.

You're right on the logging meeting we agreed NOT "to log user permissions on every api call" for security reasons.

Instead (I thought) we agreed to log the permission count and whether global-access or not, which is what this function is doing. Maybe I shouldn't have called the function log_user_permissions() then.

metrics/data/managers/core_models/time_series.py
dandammann: Of course, with pleasure!

Done.

tests/unit/common/auth/test_permissions.py
dandammann: Well-spotted. Done.

tests/unit/common/auth/test_permissions.py
dandammann: Agreed. I have now removed:

test_geography_type_match_with_geography_mismatch_denies_access
test_geography_type_mismatch_with_geography_match_denies_access

tests/unit/common/auth/test_permissions.py
dandammann: Done.

common/auth/permissions.py
dandammann: Done.

metrics/domain/models/common.py
mattjreynolds: @jrdh I think it's because in this instance it's referenced as a string - which is usually done to get round circular import issues with type checking (so no import required)

tests/unit/common/auth/test_permissions.py
dandammann: Done.

common/auth/permissions.py
dandammann: Agreed. I only kept the 2nd check now.

tests/integration/metrics/data/managers/core_models/test_time_series.py
dandammann: Agreed. I built the permission set from the non-public recordset now.

common/auth/cognito_jwt/backend.py
kathryn-dale: Log it yes, but not on every call I don't think. That feels way too noisy

dandammann: No worries, @kathryn-dale. I have removed this global log call now. I then added the log back in your cms/dashboard/viewsets.py where you had it when doing your work in the https://ukhsa.atlassian.net/browse/CDD-3171 branch.

jrdh (approved):
Looks good to me!

metrics/domain/models/common.py
jrdh: That must be an oversight on my part, I thought even if you referenced a type as a string you still needed to import it? Not important here though, I'll go figure that one out myself 😁

kathryn-dale (approved):

mattjreynolds (approved):

PR #3251, opened 2026-06-12
Enabling Entra Auth

itsthatianguy:
# Description

- When looking for an auth header, first checks for Cognito, if not present checks for Entra
- Entra token validation and user manager added
- Permissions lookup for Entra, Cognito updated to lookup permissions if not found in the token
- If the Entra App ID isn't found in the database, returns a 401

**NOTE: This is expecting the app consumer to be a created User in the database. Follow up ticket to move that into a separate entity if required**

---

## Type of change

---

# Checklist:


common/auth/jwt/backend.py
```diff
@@ -24,6 +25,12 @@ def get_authorization_header(request):
     """
     auth_header = getattr(settings, "COGNITO_JWT_AUTH_HEADER", "Authorization")
     auth = request.META.get(auth_header, b"")
+
+    # If the Cognito header isn't present, look for the Entra auth header
+    if not auth:
+        auth_header = getattr(settings, "ENTRA_JWT_AUTH_HEADER", "HTTP_AUTHORIZATION")
+        auth = request.META.get(auth_header, b"")
+
```
mattjreynolds: This section shouldn't be needed, as auth_header will default to the standard `Authorization` header if we don't provide the env variable. 
So in the private API it will use the env variable to grab the token from the `X-UHD-AUTH` header. 
In the public api (where the env variable is not set) it will grab the token from the `Authorization` header which is what we're expecting the APIM users to use.

common/auth/jwt/user_manager.py
```diff
@@ -0,0 +1,74 @@
...
+            )
+            return None
+
+        user_class = get_user_model()
+        user = user_class(username=username)
+        user.permission_sets = permission_sets
+        return user
+
+
+class EntraManager(BaseUserManager):
```
mattjreynolds, line 49: I think we can probably simplify things here by just having it try getting permission_sets from the token and fall  back to getting them locally regardless of whether it's entra or cognito.
Having said that, it might make more sense to just merge this in as-is, and we can simplify it later if we want to.
I.e the advantage of having them separated into entra and cognito classes is it makes it easier to resolve any issues we find in one without affecting the other, and then once we've ironed out any potential issues and both are working fine with real world customers we can consider a refactor to simplify at that point.

common/auth/jwt/backend.py
itsthatianguy: I wasn't sure if there would be a different header the same as with Cognito - will it definitely be the standard Authorization header?

common/auth/jwt/user_manager.py
itsthatianguy: Yeah, my thinking was both to minimize potential impact on existing stuff by adding a new manager, and also to allow a bit more flexibility going forwards if anything wants modifying for either Cognito or Entra
If we think it's unnecessary complexity though I can remove and merge into one manager?

mattjreynolds: I think I'm leaning more towards keeping it as is. 
You've done the work anyway at this point, and it makes us less likely to break one when trying to fix an issue with the other.

common/auth/jwt/backend.py
mattjreynolds: It should be the standard one, as that's not currently in use. The reason we had to make a custom one for cognito is that the private API already uses Auth header for an API key to block unwanted requests at the ALB.
Even if we do decide to use a custom header, we already have the env variable in place to provide that anyway - we just need to set it to whatever we decide.

itsthatianguy: Updated - they share the same env variable now instead 👍

mattjreynolds (approved):

kathryn-dale (approved):

PR #3255, opened 2026-06-19
Audit Logging Update

itsthatianguy:
# Description

Updates the audit logging to bring the timestamp inline with the frontend audit logging timestamp, prevents misleading logs on log in (logging in to the CMS updated a `last_login` value, which triggered the update signal), and includes both the CMS id and the UUID for the target user in the logs

---

## Type of change

---

# Checklist:


kathryn-dale (approved):

common/signals.py
```diff
@@ -117,12 +135,17 @@ def audit_delete_log(sender, instance, **kwargs):
...
+        else f"pk={instance.pk}"
+    )
 
     audit_logger.info(
         "Model deleted",
         extra={
             "user": user_id,
-            "action": f"DELETE {sender.__name__}",
-            "target": f"id={instance.pk}",
+            "action": f"pk={instance.pk}, DELETE {sender.__name__}",
```
dw-ukhsa: Including the `pk` here seems possible it could impact downstream queries that might rely on `action` being in a strict format? That aside its already in the `target_string` so is probably not necessary

itsthatianguy: Good catch - updated 👍

dw-ukhsa (approved):
All looks good :)

PR #3267, opened 2026-06-30
CDD-3440: Add regression coverage for `/api/pages/` filtering and search

dw-ukhsa:
# Description

- Add API-level integration coverage for `GET /api/pages/` filtering and search behaviour
- Covers unauthenticated search requests when `AUTH_ENABLED=True`, ensuring private pages matching the search term are excluded
- Covers `AUTH_ENABLED=False`, ensuring `/api/pages` returns the unfiltered public-prod result
- Covers authenticated restricted-user search requests, ensuring results include public and permitted private pages only.
- Refactors the unauthenticated `AUTH_ENABLED=True` filtering path to return a Wagtail-search-compatible `id__in` queryset

Fixes #CDD-3440

---

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

---

# Checklist:

# Testing:
```bash
python -m pytest tests/integration/cms/dashboard/test_viewsets.py


mattjreynolds (approved):
Looks good :sunglasses:
It needs a rebase onto latest main, and it's failing the linting CI check, you can fix this by just running the linting fix in the uhd scripts:
`source uhd.sh`
`uhd quality format`

Nice work on cleaning up the tech debt in the queryset filtering in `cms/dashboard/viewsets.py` too, that's much more readable now :raised_hands: 
I've double checked that the integration tests pass even without those improvements, but given that we are changing the code as well as the tests we should run this through QA just to get sign off from Shahrukh that we don't have any regressions @sahmed06

dw-ukhsa:
@mattjreynolds
Thanks for the heads up on the linting with `uhd` - didn't know it was a thing until now.
Linting should be fixed and rebased to current main 🦈

dw-ukhsa:
Updated the PR to reflect the feedback:

current scope is now:
- Expanded `tests/integration/cms/dashboard/test_viewsets.py` with regression coverage for `/api/pages` search and detail behaviour.
- Covered anonymous, restricted-user, global-access, and `AUTH_ENABLED=False` paths.
- Refactored `CMSPagesAPIViewSet.get_queryset()` so anonymous and authenticated filtering both build allowed page IDs and return `queryset.filter(id__in=...)`, keeping the result compatible with Wagtail search queryseys.

mattjreynolds (approved):
Looks good 😎

dw-ukhsa:
As per discussion around extra safeguards for non-pubic pages not being searchable if `AUTH_ENABLED = False`, I have changed the logic in `get_queryset()` to return public pages only when `AUTH_ENABLED is false OR the user is not authorised`.

Added an extra regression test to cover this scenario.

Will need reviewing again, did anyone say @mattjreynolds

mattjreynolds (approved):

MeganB-2 (approved):

PR #3269, opened 2026-06-30
Cdd 3451 be make sure all 6 permission fields get passed with every charts api call

dandammann:
# Description

See ticket: https://ukhsa.atlassian.net/browse/CDD-3451

---

## Type of change

- [X] Tech debt item

---

# Checklist:


metrics/data/managers/core_models/topic.py
MeganB-2: Copy pasta typo - Custom model manager class for the `Topic` model.

metrics/api/serializers/plots.py
```diff
@@ -10,6 +10,14 @@
...
+        help_text=help_texts.THEME_FIELD,
+        required=True,
+    )
+    sub_theme = serializers.CharField(
+        help_text=help_texts.SUB_THEME_FIELD,
+        required=True,
+    )
     topic = serializers.ChoiceField(
         help_text=help_texts.TOPIC_FIELD,
         choices=[],
```
MeganB-2, line 23: What is choices doing here?

metrics/data/managers/core_models/geography.py
```diff
@@ -233,6 +258,21 @@ def get_code_by_name(
...
+        """
+        Gets the geography_type id and geography code for the given names in a
+        single query that also validates the pair is consistent.
+
+        Returns:
+            A tuple of (geography_type_id, geography_code) if the pair exists,
+            or (None, None) if it does not.
+        """
+        return self.get_queryset().get_geography_type_id_and_code_by_name(
+            geography_name, geography_type_name=geography_type_name
```
MeganB-2: Don't need 'geography_type_name=' here as it's positional arguments

metrics/api/serializers/plots.py
itsthatianguy: Not sure, the docs aren't too clear:
https://www.django-rest-framework.org/api-guide/fields/#choicefield

Since they all seem to have it, and it hasn't been added, I'm wondering if you have to provide a default when initializing, and can override it with the actual values later? 🤔

metrics/data/managers/core_models/geography.py
itsthatianguy: Removed

metrics/data/managers/core_models/topic.py
itsthatianguy: I've fixed a few of those in the codebase 😅
Guessing because this wasn't added here that Dan missed it, fixed now

metrics/api/serializers/plots.py
MeganB-2: Ah yeah, my line of thinking was that we haven't added it to the new geography and geography_type serializers

itsthatianguy: Yeah, think it's just the different field types 👍

MeganB-2 (dismissed):

cms/utils/migrations/backfill_helpers.py
```diff
@@ -0,0 +1,57 @@
...
+        for item in node:
+            changed |= _fill_node(item, topic_map)
+
+    return changed
+
+
+def backfill_pages_with_theme_and_subtheme(apps, model_section, model_name):
+    topic_map = _build_topic_map(apps)
+
+    LandingPage = apps.get_model(model_section, model_name)
```
jrdh: This var is confusingly named as the function is called with

- `backfill_pages_with_theme_and_subtheme(apps, "home", "LandingPage")`
- `backfill_pages_with_theme_and_subtheme(apps, "topic", "TopicPage")`
- `backfill_pages_with_theme_and_subtheme(apps, "topics_list", "TopicsListPage")`

so sometimes it's a `LandingPage` but other times it's a `TopicPage` or a `TopicsListPage`.

Non-blocking issue, just confusing to read through.

cms/utils/migrations/backfill_helpers.py
```diff
@@ -0,0 +1,57 @@
+from wagtail.fields import StreamValue
```
jrdh, line 12: Thinking forward to someone looking at this in 6 months time with no awareness of what we're doing now, I think it'd be good to have a block comment at the top of this file explaining what these functions were used for and why.

jrdh (dismissed):
Changes look good. I've included a couple of non-blocking comments about naming and docs but the change itself looks solid to me.

cms/utils/migrations/backfill_helpers.py
itsthatianguy: Ah good catch - missed during copy and paste 🙈
`app_model` seemed to be an appropriate name

cms/utils/migrations/backfill_helpers.py
itsthatianguy: Added some info - let me know that's clear or if more/less info is needed

mattjreynolds (approved):

mattjreynolds (dismissed):

MeganB-2 (dismissed):

MeganB-2 (approved):

jrdh (approved):

PR #3294, opened 2026-08-03
fix: resolve page classification clashes [CDD-3498]

jrdh:
# Description

Topic and metric documentation child pages are non-public capable pages which therefore must have the non-public tick box, page classification, theme, sub-theme, topic, and metric associated with them to allow us to filter based on a user's permissions if the page is set to non-public.

In the original implementation of this, there was a bug where the metric field on the documentation child page model was being reused as it already existed. Conceptually this is incorrect as it should have been a separate field, but this approach also broke the metric child pages as they were expecting the metric to be stored as the full name of the metric whereas the non-public logic was expecting an ID. To solve this, the metric field has been removed because it's not currently in scope for page permissions and was only half implemented anyway.

To future proof non-public pages and also reduce repeated code (the topic and metric documentation child models have the same code copied in both) a new non-public abstract page model has been created which carries the core fields required for a non-public page. The topic and metric documentation child page models are now subclasses of this page.

Additionally, the theme, sub-theme, and topic fields have been renamed with the prefix `page_` to avoid clashing with possible fields on subclasses. This aligns with the `page_classification` that already existed.

On top of this, the admin form for non-public fields has been centralised into a class as well to again avoid copied code. This presents the fields slightly differently to try and make the DPD editor user experience better - the public checkbox still shows as before but the classification, theme, sub-theme, and topic are grouped together and collapsed by
default to reduce clutter when they are likely only to be set once and not looked at again. These could be moved into a tab potentially if we wanted to but for the moment this change is better than before and makes it clearer for users.

This means this:
image

Now looks like this:
image

There is also a performance fix within this PR which I tried to split out but couldn't as it's related to the changes described above. Specifically, the populating of choices on the metric documentation child model is moved from the `__init__` method to the admin form's `__init__`. This means we don't pull metric names from the db on every instantiation of a new metric documentation object (which for example could happen 80+ times (depends how many
metric documentation child pages you have) just through the get pages endpoint (because we use `.specific()`).

Finally, to facilitate the field changes on the topic and metric documentation child pages, migrations have been created. These are written as renames to try and avoid data loss. However, any metric documentation child pages which were created with the old code will need to fixed as their database entries will be incorrect (they'll be using an ID in the `metric` column instead of a string name). This has been fixed in the bootstrap code as well as the issue was present there too.

Fixes #CDD-3498

---

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] Tech debt item (this is focused solely on addressing any relevant technical debt)

---

# Checklist:


cms/auth_content/static/js/toggle_available_fields_on_is_public.js
```diff
@@ -170,30 +151,6 @@
     }
   }
 
-  /**
-   * Handle topic selection change
-   */
-  async function handleTopicChange() {
```
itsthatianguy, line 176: Part of what this was in place for was to narrow down the available metrics based on the selected topic - so e.g. you couldn't select a weather alert metric for a covid-19 topic
If we're happy to regress that behaviour that's fine, just wanted to call it out and check

cms/auth_content/models/non_public_page.py
```diff
@@ -0,0 +1,113 @@
...
+class NonPublicPage(models.Model):
+    """
+    An abstract model to be reused on pages which require non-public access control.
+    """
+
+    class Meta:
+        abstract = True
+
+    # whether the page should be publicly available or not
+    is_public = models.BooleanField(
```
itsthatianguy, line 30: Bit of a nitpick, but to me this makes the naming of the page class confusing - e.g. how can a `NonPublicPage` be public?

tests/unit/cms/auth_content/forms/test_non_public_page.py
```diff
@@ -0,0 +1,146 @@
...
+from cms.auth_content.models.non_public_page import (
+    DataClassificationLevels,
+    NonPublicPage,
+)
+from cms.topic.models import TopicPage
+
+
+class TestNonPublicPageAdminForm:
+
+    @pytest.mark.django_db
```
itsthatianguy: I believe usage of `@pytest.mark.django_db` means the test counts as an integration test, rather than a unit test

jrdh: Ahhh yeah fair point. We should add this as a linting step ot something.

cms/auth_content/static/js/toggle_available_fields_on_is_public.js
jrdh: I removed this because I've removed the metric level page classification so this wasn't needed any more. It was half implemented so thought it was better to remove than leave half of it in place.

jrdh: I say half implemented btw because:
- some of the metric related logic was commented out in various places (including this js file iirc)
- the topic page never had a metric attribute on it
- the metric attribute on the metrics documentation child page only existed because that page already had it and that's the bug this PR is fixing cause it shouldn't have been reusing that attribute

So I've just removed it. I believe for page level access control we only needed to go down as far as topic, hence removing it makes sense anyway to avoid confusion. We can always add it back in later if needed.

itsthatianguy (approved):

itsthatianguy (approved):

jrdh:
This PR got stale and changes were made in the same place in #3322 which clashed hard making the merge conflict challenging to resolve cleanly. This has been done on a separate branch to make it simpler to parse through and thus a new PR supersedes this one: https://github.com/UKHSA-Internal/data-dashboard-api/pull/3362.

